PR #1, opened 2020-07-21
Initialize class.h

ddacw:
**Changes:** created `class Thumbnailer` and its components.

### DIFF
--- a/src/class.h
+++ b/src/class.h
@@ -12,14 +12,45 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <algorithm>
+#include <vector>
+
+#include "../imageio/image_dec.h"
+#include "../imageio/imageio_util.h"
 #include "webp/encode.h"
+#include "webp/mux.h"
 
 namespace libwebp {
 
-// Class that does something.
+// Takes time stamped images as an input and produces an animation.
 class Thumbnailer {
-  // TODO: add an API to add frames with a timestamp.
-  // TODO: add an API to generate the animation.
+ public:
+  Thumbnailer();
+  ~Thumbnailer();
+
+  // Adds a frame with a timestamp (in millisecond).
+  // Returns true on success and false on failure.
+  bool AddFrame(const WebPPicture& pic, int timestamp_ms);
+
+  // Generates the animation.
+  // Returns true on success and false on failure.
+  bool GenerateAnimation(WebPPicture* const output);
+
+ private:
+  struct FrameData {
+    WebPPicture pic;
+    int timestamp_ms;
+  };
+  std::vector<FrameData> frames;
+  WebPAnimEncoder* enc = NULL;
+  WebPData webp_data;
+  WebPAnimEncoderOptions anim_config;
+  WebPConfig config;
+  int loop_count;
 };
 
 }  // namespace libwebp


src/class.h
```diff
@@ -12,14 +12,39 @@
...
+#include "webp/mux.h"
 
 namespace libwebp {
 
-// Class that does something.
+// Takes time stamped images as an input and produces an animation.
 class Thumbnailer {
-  // TODO: add an API to add frames with a timestamp.
-  // TODO: add an API to generate the animation.
+  WebPAnimEncoder* enc = NULL;
```
vrabaud: Make the members private.

src/class.h
```diff
@@ -12,14 +12,39 @@
...
+  int timestamp = 0;
+  int width;
+  int height;
+  int duration = 500;
+  int loop_count;
+  int num_pic = 0;
+
+  // Initialize
+  Thumbnailer();
+
```
vrabaud, line 34: You will need a destructor to destruct WebPAnimEncoder (or wrap it in a unique_ptr)

src/class.h
```diff
@@ -12,14 +12,39 @@
...
+  // Add frame with a timestamp
+  int AddFrame(const char* filename);
```
vrabaud: how do you add the timestamp ?
What is the int that you return ?

src/class.h
```diff
@@ -12,14 +12,39 @@
...
+  int width;
+  int height;
+  int duration = 500;
+  int loop_count;
+  int num_pic = 0;
+
+  // Initialize
+  Thumbnailer();
+
+  // Add frame with a timestamp
```
vrabaud: When starting a comment with capital letter, please end it with a dot:   '.'

src/class.h
```diff
@@ -12,14 +12,39 @@
...
+  int AddFrame(const char* filename);
+
+  // Generate the animation
+  int GenerateAnimation(const char* output);
```
vrabaud: what is the returned int ? If it is a status code (please make it an enum), what can happen ?

src/class.h
vrabaud: Ideally, this class would not deal with the filesystem: it should receive an image buffer (as uint8_t*) and/or a WebPPicture

vrabaud (commented):
Good start !

src/class.h
```diff
@@ -12,14 +12,40 @@
...
+  std::vector<WebPPicture> pics;
+  WebPData webp_data;
+  WebPAnimEncoderOptions anim_config;
+  WebPConfig config;
+  std::vector<int> timestamps;
+  int loop_count;
+  int num_pic = 0;
+
+ public:
+  Thumbnailer();
```
vrabaud, line 32: Please put the public members first (as mentioned in https://google.github.io/styleguide/cppguide.html)

src/class.h
```diff
@@ -12,14 +12,40 @@
...
-  // TODO: add an API to generate the animation.
+ private:
+  WebPAnimEncoder* enc = NULL;
+  std::vector<WebPPicture> pics;
+  WebPData webp_data;
+  WebPAnimEncoderOptions anim_config;
+  WebPConfig config;
+  std::vector<int> timestamps;
+  int loop_count;
+  int num_pic = 0;
```
vrabaud: Is that still useful ? Probably redundant with timestamps.size().

src/class.h
```diff
@@ -12,14 +12,40 @@
...
 class Thumbnailer {
-  // TODO: add an API to add frames with a timestamp.
-  // TODO: add an API to generate the animation.
+ private:
+  WebPAnimEncoder* enc = NULL;
+  std::vector<WebPPicture> pics;
+  WebPData webp_data;
+  WebPAnimEncoderOptions anim_config;
+  WebPConfig config;
+  std::vector<int> timestamps;
```
vrabaud: instead of going with two vectors, I'd create a private struct:
struct FrameData {
    WebPPicture pic;
    int timestamp_ms;
  };
And have one vector of those (always better to keep the data together)

src/class.h
```diff
@@ -12,14 +12,40 @@
...
+  int loop_count;
+  int num_pic = 0;
+
+ public:
+  Thumbnailer();
+  ~Thumbnailer();
+
+  // Add frame with a timestamp.
```
vrabaud: what is the unit of the timestamp ?

src/class.h
```diff
@@ -12,14 +12,40 @@
...
+  bool AddFrame(const WebPPicture& pic, int timestamp);
```
vrabaud: What does the boolean mean ?

src/class.h
```diff
@@ -12,14 +12,40 @@
...
+
+  // Generate the animation.
+  bool GenerateAnimation(WebPPicture* const output);
```
vrabaud, line 41: What does the boolean mean ?
(simply write something like: Returns true on success and false on failure)

src/class.h
```diff
@@ -12,14 +12,40 @@
...
+  int num_pic = 0;
+
+ public:
+  Thumbnailer();
+  ~Thumbnailer();
+
+  // Add frame with a timestamp.
+  bool AddFrame(const WebPPicture& pic, int timestamp);
+
+  // Generate the animation.
```
vrabaud: Please conjugate action in description (it is a Google style): Generates the animation.
(same above with "Adds frame").

src/class.h
ddacw: What should I write if both are the same? (true on success and false on failure)

vrabaud:
Please squash your commits and we are good to go !

PR #8, opened 2020-07-22
Implement class.cc and modify class.h

uyennguyen0903:
Implemented class.cc and added define guard to class.h

### DIFF
--- a/src/class.cc
+++ b/src/class.cc
@@ -12,14 +12,83 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "class.h"
+
 namespace libwebp {
 
-// Let's implement stuff !
-// TODO: to create an animation:
-//       - create a WebPAnimEncoderOptions  config
-//       - set the quality to something in your search space
-//       - call WebPAnimEncoderNew, then fill the animation with WebPPicture
-//         using WebPAnimEncoderAdd
-//       - finalize with WebPAnimEncoderAssemble
+Thumbnailer::Thumbnailer() {
+  WebPAnimEncoderOptionsInit(&anim_config);
+  WebPConfigInit(&config);
+}
+
+Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc); }
+
+Thumbnailer::Status Thumbnailer::AddFrame(const WebPPicture& pic,
+                                          int timestamp_ms) {
+  // Verify dimension of frames.
+  if (frames.empty()) {
+    enc = WebPAnimEncoderNew(pic.width, pic.height, &anim_config);
+    if (enc == nullptr) {
+      return kMemoryError;
+    }
+  } else {
+    if (pic.width != (frames[0].pic).width ||
+        pic.height != (frames[0].pic).height) {
+      return kImageFormatError;
+    }
+  }
+
+  FrameData new_frame = {pic, timestamp_ms};
+  frames.push_back(new_frame);
+
+  return kOk;
+}
+
+Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
+  bool ok = true;
+
+  // Rearrange frames.
+  std::sort(frames.begin(), frames.end(),
+            [](const FrameData& A, const FrameData& B) -> bool {
+              return A.timestamp_ms < B.timestamp_ms;
+            });
+
+  // Fill the animation.
+  for (auto& frame : frames) {
+    ok = ok && WebPAnimEncoderAdd(enc, &frame.pic, frame.timestamp_ms, &config);
+  }
+
+  // Add last frame.
+  ok = ok && WebPAnimEncoderAdd(enc, NULL, frames.back().timestamp_ms, NULL);
+  ok = ok && WebPAnimEncoderAssemble(enc, webp_data);
+
+  // Set loop count.
+  WebPMuxError err;
+  WebPMuxAnimParams new_params;
+  uint32_t features;
+
+  std::unique_ptr<WebPMux, void (*)(WebPMux*)> mux(WebPMuxCreate(webp_data, 1),
+                                                   WebPMuxDelete);
+
+  if (mux.get() == nullptr) return kMemoryError;
+
+  if (WebPMuxGetFeatures(mux.get(), &features) != WEBP_MUX_OK) {
+    return kMemoryError;
+  }
+
+  if (WebPMuxGetAnimationParams(mux.get(), &new_params) != WEBP_MUX_OK) {
+    return kMemoryError;
+  }
+
+  new_params.loop_count = loop_count;
+  if (WebPMuxSetAnimationParams(mux.get(), &new_params) != WEBP_MUX_OK) {
+    return kMemoryError;
+  }
+
+  WebPDataClear(webp_data);
+  if (WebPMuxAssemble(mux.get(), webp_data) != WEBP_MUX_OK) return kMemoryError;
+
+  return kOk;
+}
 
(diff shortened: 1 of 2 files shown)


src/class.cc
```diff
@@ -12,14 +12,65 @@
...
+    }
+  }
+
+  FrameData new_frame = {pic, timestamp_ms};
+  frames.push_back(new_frame);
+
+  return ok;
+}
+
+bool Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
```
ddacw: TODO: catch errors.

src/class.cc
```diff
@@ -12,14 +12,65 @@
...
+              return A.timestamp_ms < B.timestamp_ms;
+            });
+
+  // Fills the animation.
+  for (auto frame : frames) {
+    ok = ok && WebPAnimEncoderAdd(enc, &frame.pic, frame.timestamp_ms, &config);
+  }
+
+  // Adds last frame.
+  WebPAnimEncoderAdd(enc, NULL, frames.back().timestamp_ms, NULL);
```
uyennguyen0903: Check if (ok == true) before adding the last frame.

src/class.cc
```diff
@@ -12,14 +12,65 @@
...
+  bool ok = true;
+
+  // Verifies dimension of frames.
+  if (frames.empty()) {
+    enc = WebPAnimEncoderNew(pic.width, pic.height, &anim_config);
+    ok = (enc != nullptr);
+  } else {
+    if (pic.width != (frames[0].pic).width ||
+        pic.height != (frames[0].pic).height) {
+      ok = false;
```
vrabaud: mayeb return an enum instead. Create one that has: kMemoryError, kImageFormatError, and kOk

src/class.cc
```diff
@@ -12,14 +12,65 @@
...
+
+  return ok;
+}
+
+bool Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
+  bool ok = true;
+
+  // Rearranges frames.
+  std::sort(frames.begin(), frames.end(),
+            [&](FrameData A, FrameData B) -> bool {
```
vrabaud: const FrameData& A to avoid memory copies.
[] is enough as you do not capture anything.

src/class.cc
```diff
@@ -12,14 +12,65 @@
...
+              return A.timestamp_ms < B.timestamp_ms;
+            });
+
+  // Fills the animation.
+  for (auto frame : frames) {
```
vrabaud: const auto& frame

src/class.cc
```diff
@@ -12,14 +12,65 @@
...
+    ok = ok && WebPAnimEncoderAdd(enc, &frame.pic, frame.timestamp_ms, &config);
```
vrabaud: ok &= WebPAnim ...

vrabaud (commented):
ok, looks good overall !

src/class.cc
```diff
@@ -12,14 +12,65 @@
...
+  if (frames.empty()) {
+    enc = WebPAnimEncoderNew(pic.width, pic.height, &anim_config);
+    ok = (enc != nullptr);
+  } else {
+    if (pic.width != (frames[0].pic).width ||
+        pic.height != (frames[0].pic).height) {
+      ok = false;
+    }
+  }
+
```
skal65535, line 40:
```suggestion
if (!ok) return false;
```
(that is: don't add the frame if something went wrong before)

src/class.cc
```diff
@@ -12,14 +12,65 @@
...
+    ok = ok && WebPAnimEncoderAdd(enc, &frame.pic, frame.timestamp_ms, &config);
+  }
+
+  // Adds last frame.
+  WebPAnimEncoderAdd(enc, NULL, frames.back().timestamp_ms, NULL);
+  WebPAnimEncoderAssemble(enc, webp_data);
+
+  // Sets loop count.
+  WebPMuxAnimParams new_params;
+  WebPMux* const mux = WebPMuxCreate(webp_data, 1);
```
skal65535: mux can be returned NULL in case of memory allocation failure. Catch this case.


I'd suggest catching errors right away went writing the initial code instead of leaving it as a TODO for later.
There's no real value in delaying the error-catching code. It *has* to be done, so better sooner than later, and be done with it.

src/class.h
```diff
@@ -32,13 +36,19 @@ class Thumbnailer {
...
   ~Thumbnailer();
 
+  // Error codes for adding frame.
+  enum AddFrameError { kOk = 0, kMemoryError, kImageFormatError };
+
   // Adds a frame with a timestamp (in millisecond).
-  // Returns true on success and false on failure.
-  bool AddFrame(const WebPPicture& pic, int timestamp_ms);
+  // Returns:
+  //   kOk - on success.
```
vrabaud: Please move those three comments in the enum definition:
kOk=0,               // on success
kMemoryError,   // in case of memory error
And for this function, no need to explain the return argument, we know it is a status.

src/class.h
```diff
@@ -32,13 +36,19 @@ class Thumbnailer {
   Thumbnailer();
   ~Thumbnailer();
 
+  // Error codes for adding frame.
+  enum AddFrameError { kOk = 0, kMemoryError, kImageFormatError };
```
vrabaud: Please rename AddFrameError to Status (if it is kOk, it is not an error).

src/class.cc
vrabaud: The best thing to do is encapsulate the deleter in a unique_ptr:
https://en.cppreference.com/w/cpp/memory/unique_ptr
std::unique_ptr<WebPMux*> mux(WebPMuxCreate(webp_data, 1), WebPMuxDelete);

This way, you can just check if it is null. If so, return kMemoryError right away.
Otherwise, you do not need to call WebPMuxDelete at the end, it will be called automatically.
And you do not need the if (ok) below. If you have an error, you can return right away.

src/class.cc
```diff
@@ -12,14 +12,89 @@
...
+
+  // Rearranges frames.
+  std::sort(frames.begin(), frames.end(),
+            [](const FrameData& A, const FrameData& B) -> bool {
+              return A.timestamp_ms < B.timestamp_ms;
+            });
+
+  // Fills the animation.
+  for (auto& frame : frames) {
+    if (ok)
```
vrabaud: if (ok) {
}

src/class.cc
```diff
@@ -12,14 +12,83 @@
...
+            [](const FrameData& A, const FrameData& B) -> bool {
+              return A.timestamp_ms < B.timestamp_ms;
+            });
+
+  // Fills the animation.
+  for (auto& frame : frames) {
+    ok = ok && WebPAnimEncoderAdd(enc, &frame.pic, frame.timestamp_ms, &config);
+  }
+
+  // Adds last frame.
```
vrabaud: for a comment, please keep direct style: Add last frame.
The Adds last frame is just for a comment for a function. I know ...

src/class.h
```diff
@@ -47,10 +57,11 @@ class Thumbnailer {
...
-  WebPData webp_data;
   WebPAnimEncoderOptions anim_config;
   WebPConfig config;
-  int loop_count;
+  int loop_count = 0;
 };
 
 }  // namespace libwebp
+
+#endif  // THUMBNAILER_SRC_CLASS_H_
```
vrabaud, line 67: Please make sure your file has an extra line at the end

ddacw: I think we do have an extra line but it doesn't show up here.

vrabaud:
Awesome !! Great start !

PR #9, opened 2020-07-24
Implement main.cc

ddacw:


CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ project(
   VERSION 0.1.0)
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_CXX_STANDARD 17)
+add_compile_definitions(WEBP_HAVE_PNG)
```
vrabaud: This line is not needed anymore: please rebase your branch to get the image library fixes.

src/main.cc
```diff
@@ -12,14 +12,55 @@
...
+#include <string>
 
 #include "class.h"
 
-int main(int argc, char *argv[]) {
+// Return true on success or false on failure.
+static bool ReadImage(const char filename[], WebPPicture* const pic) {
+  const uint8_t* data = NULL;
+  size_t data_size = 0;
+  WebPImageReader reader;
```
vrabaud: You do not need to define all your variables at the beginning of the file: declare them whenever you need them: it makes the code shorter, it helps the compiler and prevents the person who codes to back back and forth.
E.g., below, please write: WebPImageReader reader = WebPGuessImageReader(data, data_size);

src/main.cc
```diff
@@ -12,14 +12,55 @@
...
+
+  std::ifstream input_list(argv[1]);
+  const char* output = argv[2];
+  std::string filename_str;
+  int timestamp_ms;
+
+  while (input_list >> filename_str >> timestamp_ms) {
+    char filename[filename_str.length()];
+    strcpy(filename, filename_str.c_str());
+    ReadImage(filename, &pic);
```
vrabaud: You are using the same picture over and over again thus re-writing the memory inside. You need to create a new pic every time thus having a vector of WebPPicture in main.cc. I know it is weird but the main thread should keep all the images in memory until GenerateAnimation is called.
This is because we want to prevent too many image copies over the place (it is slow ....)

src/main.cc
```diff
@@ -12,14 +12,55 @@
...
+    thumbnailer.AddFrame(pic, timestamp_ms);
+  }
+
+  // thumbnailer.GenerateAnimation(&webp_data); // currently SEGFAULT
```
vrabaud: Please uncomment :)

src/main.cc
```diff
@@ -12,14 +12,55 @@
...
+  if (!ImgIoUtilReadFile(filename, &data, &data_size)) return false;
+  reader = WebPGuessImageReader(data, data_size);
+  ok = reader(data, data_size, pic, 1, NULL);
+
+  free((void*)data);
+  return ok;
+}
+
+int main(int argc, char* argv[]) {
   // TODO: read images given as arguments
```
vrabaud: Those can go now !

vrabaud:
you still need to keep the other changes you had done to CMakeLists.txt. You need main.cc to be linker to imageio_utils for example.

src/main.cc
vrabaud: You now have a memory leak: your pics are created but never freed. Store them in a vector of unique_ptr so that they are magically deleted at the end.

src/main.cc
```diff
@@ -12,14 +12,50 @@
...
+int main(int argc, char* argv[]) {
+  libwebp::Thumbnailer thumbnailer = libwebp::Thumbnailer();
+
+  WebPData webp_data;
+  WebPDataInit(&webp_data);
+
+  std::ifstream input_list(argv[1]);
+  const char* output = argv[2];
+  std::string filename_str;
+  int timestamp_ms;
```
vrabaud, line 41: please declare those in the for loop (usually declare variables in the smallest scope possible).

ddacw: Does that mean I should use a `while (true)` loop for this case?

PR #11, opened 2020-07-28
Use binary search to find quality factor.

uyennguyen0903:
Added binary search algorithm to API to find the best quality that makes the animation fit right below the given byte budget.


vrabaud:
Nice ! That does the trick !

PR #14, opened 2020-07-30
Add Gtest

ddacw:
Add testing and update CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,4 +29,4 @@ jobs:
     
     # Run tests
     - name: Run tests
-      run: valgrind ./build/test/test_thumbnailer
+      run: valgrind --leak-check=full --show-leak-kinds=all ./build/test/test_thumbnailer
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,18 +11,22 @@ jobs:
   # This workflow contains a single job called "build"
   build:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - uses: actions/checkout@v2
 
-    # Install dependencies.
+    # Install dependencies
     - name: Install dependencies
-      run: sudo apt install cmake g++ libjpeg-dev libpng-dev libprotobuf-dev libwebp-dev ninja-build protobuf-compiler
+      run: sudo apt install cmake g++ libjpeg-dev libpng-dev libprotobuf-dev libwebp-dev libgtest-dev ninja-build protobuf-compiler valgrind
 
     # Build the code
     - name: Build
       run: |
-        mkdir build && cd build && cmake -G "Ninja" ../ && ninja
+        mkdir build && cd build && cmake -G "Ninja" ../ && ninja 
+    
+    # Run tests
+    - name: Run tests
+      run: valgrind ./build/test/test_thumbnailer
--- a/test/main.cc
+++ b/test/main.cc
@@ -1,0 +1,115 @@
+#include "../src/class.h"
+#include "gtest/gtest.h"
+
+// Returns RGBA values for WebPPicture.
+// R, G, B values can be randomized or set to a common fixed value.
+std::vector<uint8_t> GenerateRGBA(int width, int height, bool randomized,
+                                  int fixed, uint8_t transparency) {
+  std::vector<uint8_t> rgba;
+  for (int i = 0; i <= height; ++i) {
+    for (int j = 0; j <= width * 4; ++j) {
+      for (int k = 0; k < 3; ++k) {
+        rgba.push_back(randomized ? rand() & 0xff : fixed);
+      }
+      rgba.push_back(transparency);
+    }
+  }
+  return rgba;
+}
+
+// Returns vector of WebPPicture.
+std::vector<std::unique_ptr<WebPPicture, void (*)(WebPPicture*)>> GeneratePics(
+    int pic_count, int width, int height, bool randomized = 0, int fixed = 0xff,
+    uint8_t transparency = 0xff) {
+  std::vector<std::unique_ptr<WebPPicture, void (*)(WebPPicture*)>> pics;
+  for (int i = 0; i < pic_count; ++i) {
+    pics.emplace_back(new WebPPicture, WebPPictureFree);
+    WebPPictureInit(pics.back().get());
+    pics.back().get()->use_argb = 1;
+    pics.back().get()->width = width;
+    pics.back().get()->height = height;
+    WebPPictureImportRGBA(
+        pics.back().get(),
+        GenerateRGBA(width, height, randomized, fixed, transparency).data(),
+        width * 4);
+  }
+  return pics;
+}
+
+TEST(blank_image_solid, blank_images) {
+  libwebp::Thumbnailer thumbnailer = libwebp::Thumbnailer();
+
+  int pic_count = 10;
+  int width = 50;
+  int height = 50;
+  auto test_pics = GeneratePics(pic_count, width, height, 0, 0xaf);
+
+  for (int i = 0; i < pic_count; ++i) {
+    thumbnailer.AddFrame(*test_pics[i].get(), i * 500);
+  }
+  std::unique_ptr<WebPData, void (*)(WebPData*)> webp_data(new WebPData,
+                                                           WebPDataClear);
+  WebPDataInit(webp_data.get());
+  thumbnailer.GenerateAnimation(webp_data.get());
+  EXPECT_LE(webp_data.get()->size, 153600);
+}
+
+TEST(blank_image_transparent, blank_images) {
+  libwebp::Thumbnailer thumbnailer = libwebp::Thumbnailer();
+
+  int pic_count = 10;
+  int width = 50;
+  int height = 50;
+  auto test_pics = GeneratePics(pic_count, width, height, 0, 0xff, 0xaf);
+
+  for (int i = 0; i < pic_count; ++i) {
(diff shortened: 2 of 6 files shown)


test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+  // Returns RGBA values for WebPPicture.
+  std::vector<uint8_t> GenerateRGBA(int seed) {
+    std::mt19937 rng(seed);
+    std::vector<uint8_t> rgba;
+
+    int color_R = rng();
+    int color_G = rng();
+    int color_B = rng();
+
+    for (int i = 0; i <= height; ++i) {
```
miwucs: you have an off by one, this should be < height, and same for width below

test/main.cc
```diff
@@ -0,0 +1,151 @@
+#include <random>
+
+#include "../src/class.h"
+#include "gtest/gtest.h"
+
+const int kDefaultWidth = 80;
+const int kDefaultHeight = 60;
+
+class WebPSamplesGenerator {
```
miwucs: I'm not sure this really needed to be a class but it works so you can leave as is if you want

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+  WebPSamplesGenerator(int pic_count, int width, int height,
+                       uint8_t transparency, bool randomized)
+      : pic_count(pic_count),
+        width(width),
+        height(height),
+        transparency(transparency),
+        randomized(randomized) {}
+
+  // Returns RGBA values for WebPPicture.
+  std::vector<uint8_t> GenerateRGBA(int seed) {
```
miwucs: this could be private

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+      pic.get()->use_argb = 1;
+      pic.get()->width = width;
+      pic.get()->height = height;
+      WebPPictureImportRGBA(pic.get(), GenerateRGBA(i).data(), width * 4);
+    }
+    return pics;
+  }
+
+ private:
+  int pic_count;
```
miwucs: if we're following the google c++ style guide (https://google.github.io/styleguide/cppguide.html#Variable_Names), class members should end with an underscore, so pic_count_, width_ and so on.

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+    std::mt19937 rng(seed);
+    std::vector<uint8_t> rgba;
+
+    int color_R = rng();
+    int color_G = rng();
+    int color_B = rng();
+
+    for (int i = 0; i <= height; ++i) {
+      for (int j = 0; j <= width * 4; ++j) {
+        rgba.push_back(randomized ? rng() & 0xff : color_R);
```
miwucs: you also need &0xff for color_R, no? something like:
(randomized ? rng() : color_R) & 0xff

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+        height(height),
+        transparency(transparency),
+        randomized(randomized) {}
+
+  // Returns RGBA values for WebPPicture.
+  std::vector<uint8_t> GenerateRGBA(int seed) {
+    std::mt19937 rng(seed);
+    std::vector<uint8_t> rgba;
+
+    int color_R = rng();
```
miwucs: these can be const (we usually like to use const on all variables that are only assigned once)

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+  std::vector<std::unique_ptr<WebPPicture, void (*)(WebPPicture*)>>
+  GeneratePics() {
+    std::vector<std::unique_ptr<WebPPicture, void (*)(WebPPicture*)>> pics;
+    for (int i = 0; i < pic_count; ++i) {
+      pics.emplace_back(new WebPPicture, WebPPictureFree);
+      auto& pic = pics.back();
+      WebPPictureInit(pic.get());
+      pic.get()->use_argb = 1;
+      pic.get()->width = width;
+      pic.get()->height = height;
```
miwucs: you don't need .get() here, you can use -> directly on a unique_ptr

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+  int height;
+  bool randomized;
+  uint8_t transparency;
+};
+
+TEST(blank_image_solid, blank_images) {
+  libwebp::Thumbnailer thumbnailer = libwebp::Thumbnailer();
+
+  int pic_count = 10;
+  WebPSamplesGenerator Generator = WebPSamplesGenerator(pic_count, 0xff, false);
```
miwucs: variable names should be in lowercase ("generator")

miwucs: when some function arguments are not obvious, like "0xaf" or "false", we like to add a comment with the argument name, like this: /*
see https://google.github.io/styleguide/cppguide.html#Function_Argument_Comments

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+
+ private:
+  int pic_count;
+  int width;
+  int height;
+  bool randomized;
+  uint8_t transparency;
+};
+
+TEST(blank_image_solid, blank_images) {
```
miwucs: the usual format for test names is
TEST(TestSuiteName, TestName)
both the test suite and test name and in CamelCase. The test suite name is usually the name of the class you're testing followed by the word "Test". For example:
TEST(ThumbnailerTest, BlankImageSolid)
By the way, the test file itself should be called thumbnailer_test.cc, and the thumbnailer files should be called thumbnailer.h/cc instead of class.h/cc (I know Vincent created those files but I think it would be good to rename them in a different commit)

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+  auto test_pics = Generator.GeneratePics();
+
+  for (int i = 0; i < pic_count; ++i) {
+    thumbnailer.AddFrame(*test_pics[i].get(), i * 500);
+  }
+  std::unique_ptr<WebPData, void (*)(WebPData*)> webp_data(new WebPData,
+                                                           WebPDataClear);
+  WebPDataInit(webp_data.get());
+  thumbnailer.GenerateAnimation(webp_data.get());
+  EXPECT_LE(webp_data.get()->size, 153600);
```
miwucs: It's not clear where the value 153600 comes from. It would be good to test also with a value explicitly set in the thumbnailer options.
Could you also test that the size is greater than 0?

.github/workflows/main.yml
```diff
@@ -11,18 +11,22 @@ jobs:
...
 
     # Build the code
     - name: Build
       run: |
-        mkdir build && cd build && cmake -G "Ninja" ../ && ninja
+        mkdir build && cd build && cmake -G "Ninja" ../ && ninja 
+    
+    # Run tests
+    - name: Run tests
+      run: valgrind --leak-check=full --show-leak-kinds=all ./build/test/test_thumbnailer
```
miwucs, line 32: So you are right that there are a bunch of memory leaks! Without any options, valgrind only shows a summary of leaks with the number of blocks lost (definitely lost, indirectly lost...). This is already an error! If there are no leaks it shows "All heap blocks were freed -- no leaks are possible" instead. The flags allow showing extra info to help debugging.

I found two memory leaks in the test file, but there are also leaks in the Thumbnailer class. Those can be fixed in another commit.
- In class.cc on line 56, a new encoder is created every time GenerateAnimationNoBudget is called, but only the last one gets deleted in the destructor. Either we should make sure to only create one encoder, or the previous one must be deleted before creating a new one.
- There are also issues with the life cycle of webp_data/new_webp_data. On line 132, we are only copying the data pointer, not the actual data, so it probably gets deleted later on line 103? and I think there are WebPDataClear calls missing for webp_data and/or webp_new_data. I haven't looked into this fully so it would be great if one of you two could look into fixing it properly.

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  // Returns vector of WebPPicture(s).
+  // randomized == true: picture is random noise
+  // randomized == false: picture is random solid color
+  std::vector<std::unique_ptr<WebPPicture, void (*)(WebPPicture*)>>
+  GeneratePics() {
+    std::vector<std::unique_ptr<WebPPicture, void (*)(WebPPicture*)>> pics;
+    for (int i = 0; i < pic_count; ++i) {
+      pics.emplace_back(new WebPPicture, WebPPictureFree);
```
miwucs: This creates a memory leak because WebPPictureFree frees the WebPPicture's internal data, but it doesn't actually free the WebPPicture pointer. If you want to keep this approach, you could create a small function that calls WebPPictureFree then deletes the pointer.

test/main.cc
```diff
@@ -0,0 +1,151 @@
...
+
+  int pic_count = 10;
+  WebPSamplesGenerator Generator = WebPSamplesGenerator(pic_count, 0xaf, false);
+  auto test_pics = Generator.GeneratePics();
+
+  for (int i = 0; i < pic_count; ++i) {
+    thumbnailer.AddFrame(*test_pics[i].get(), i * 500);
+  }
+  std::unique_ptr<WebPData, void (*)(WebPData*)> webp_data(new WebPData,
+                                                           WebPDataClear);
```
miwucs: Same thing here, this is also a memory leak as WebPDataClear doesn't delete the pointer.

miwucs (changes requested):

test/main.cc
ddacw: @miwucs it's also `width` instead of `width * 4`, isn't it?

miwucs: Ha yes indeed!

test/main.cc
ddacw: I somehow thought it was (TestCaseName, TestName)....

test/main.cc
ddacw: This is because I want the color to be different on each frame (so that I know it's an animation).

miwucs: Yes it doesn't have to be a file constant, but it can still be a const variable within the function, if it's only assigned once.. The same goes for other variables like pic_count, transparency, etc. If a variable is only assigned once, it can be marked const.

miwucs (approved):

PR #15, opened 2020-08-05
Fix memory leaks in class.cc

uyennguyen0903:
Fix memory leaks in class.cc.


miwucs (approved):
I'm still wondering if there is an issue with the copy of new_webp_data to webp and the life cycle there, but if this works according to vilgrind let's go with it, and we can fix remaining issues later if there are any.

PR #16, opened 2020-08-07
Fix class.cc

uyennguyen0903:
Fixed error in class.cc

### DIFF
--- a/src/class.cc
+++ b/src/class.cc
@@ -129,7 +129,7 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
     int middle = (min_quality + max_quality) / 2;
 
     config_.quality = middle;
-    WebPDataClear(webp_data);
+    WebPDataClear(&new_webp_data);
     GenerateAnimationNoBudget(&new_webp_data);
 
     if (new_webp_data.size <= byte_budget_) {
--- a/src/class.cc
+++ b/src/class.cc
@@ -121,23 +121,24 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
   // below the given byte budget.
   int min_quality = minimum_lossy_quality_;
   int max_quality = 100;
+  int middle;
   int final_quality = -1;
   WebPData new_webp_data;
   WebPDataInit(&new_webp_data);
 
   while (min_quality <= max_quality) {
-    int middle = (min_quality + max_quality) / 2;
-
+    middle = (min_quality + max_quality) / 2;
     config_.quality = middle;
-    WebPDataClear(&new_webp_data);
     GenerateAnimationNoBudget(&new_webp_data);
 
     if (new_webp_data.size <= byte_budget_) {
       final_quality = middle;
+      WebPDataClear(webp_data);
       *webp_data = new_webp_data;
       min_quality = middle + 1;
     } else {
       max_quality = middle - 1;
+      WebPDataClear(&new_webp_data);
     }
   }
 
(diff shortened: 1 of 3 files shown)


miwucs (approved):
As I mentioned, if it's not too complicated it would be nice to add a test that would have caught this bug (test fails without the fix, and passes with the fix). It could be done in a different commit.

src/class.cc
```diff
@@ -129,7 +129,7 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
     int middle = (min_quality + max_quality) / 2;
 
     config_.quality = middle;
-    WebPDataClear(webp_data);
+    WebPDataClear(&new_webp_data);
```
y-guyon: I think there is an issue. `WebPDataClear()` must be called on each `WebPData` generated by `GenerateAnimationNoBudget()` except the one that will be output as `webp_data`.

With this fix, all `WebPData` except the last one will be cleared, which would work only if `new_webp_data.size <= byte_budget_` is true during the last `while (min_quality <= max_quality)` loop.

As a reminder:

- `WebPDataInit()` resets the `WebPData` object to an empty state, without freeing any memory,
- `WebPDataClear()` frees the memory referenced by the `WebPData` object,
- Copying a `WebPData` (with operator `=`) adds just a reference to the pointed `bytes` and does not deallocate nor allocate any memory; the previous content of the copy is leaked if not cleared beforehand and only the original or the copy must be cleared afterwards. If either the original or the copy is cleared, the other one becomes invalid (points to no longer existing memory). To deeply duplicate the data and allocate another `WebPData` object, use `WebPDataCopy()` instead (not needed here though).

y-guyon (changes requested):

src/class.cc
```diff
@@ -121,23 +121,24 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
   // below the given byte budget.
   int min_quality = minimum_lossy_quality_;
   int max_quality = 100;
+  int middle;
```
y-guyon: I do not see why you need to declare this outside of the loop.

src/class.cc
```diff
@@ -121,23 +121,24 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
-    int middle = (min_quality + max_quality) / 2;
-
+    middle = (min_quality + max_quality) / 2;
     config_.quality = middle;
-    WebPDataClear(&new_webp_data);
     GenerateAnimationNoBudget(&new_webp_data);
 
     if (new_webp_data.size <= byte_budget_) {
       final_quality = middle;
+      WebPDataClear(webp_data);
```
y-guyon, line 135: Note: It would be safer to mention in the comment of `GenerateAnimation()` in class.h that `webp_data` is expected to be initialized (otherwise `WebPDataClear()` might free some random memory somewhere because the pointer is undefined).

y-guyon (changes requested):

src/class.h
```diff
@@ -48,14 +48,17 @@ class Thumbnailer {
...
+  // outlive the last GenerateAnimation() call.
   Status AddFrame(const WebPPicture& pic, int timestamp_ms);
 
   // Generates the animation.
   Status GenerateAnimationNoBudget(WebPData* const webp_data);
 
   // Finds the best quality that makes the animation fit right below the given
-  // byte budget and generates the animation.
+  // byte budget and generates the animation. The 'webp_data' argument is
+  // expected to be initialized (otherwise WebPDataClear() might free some
```
y-guyon, line 60: The explanation inside the parenthesis is a bit too heavy for a method comment, it could have been omitted.

y-guyon (approved):

PR #17, opened 2020-08-11
Generate animation having frames with similar PSNR values. Misc fixes.

ddacw:


CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ include_directories(${Protobuf_INCLUDE_DIRS})
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 protobuf_generate_cpp(PROTO_SRCS PROTO_HDRS ${CMAKE_CURRENT_SOURCE_DIR}/src/thumbnailer.proto)
 
-add_library(thumbnailer_lib ${CMAKE_CURRENT_SOURCE_DIR}/src/class.cc ${PROTO_SRCS})
+add_library(thumbnailer_lib ${CMAKE_CURRENT_SOURCE_DIR}/src/thumbnailer.cc ${PROTO_SRCS})
```
miwucs, line 53: Thanks for renaming it!

src/main.cc
```diff
@@ -56,7 +56,12 @@ int main(int argc, char* argv[]) {
   const char* output = argv[2];
   WebPData webp_data;
   WebPDataInit(&webp_data);
-  thumbnailer.GenerateAnimation(&webp_data);
+  if (argc >= 4 && !strcmp(argv[3], "psnr")) {
+    std::cerr << "nope";
```
miwucs: remove :)

src/thumbnailer.cc
```diff
@@ -150,4 +146,119 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
   return kOk;
 }
 
+Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
+    WebPData* const webp_data) {
+  // Rearrange frames.
+  std::sort(frames_.begin(), frames_.end(),
+            [](const FrameData& A, const FrameData& B) -> bool {
+              return A.timestamp_ms < B.timestamp_ms;
+            });
```
miwucs: this can be moved to a helper function since it's also used in GenerateAnimation

src/thumbnailer.cc
```diff
@@ -150,4 +146,119 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+        int final_quality = -1;
+
+        // Get lowest PSNR value possible.
+        WebPPictureInit(&pic);
+        WebPPictureCopy(&frame.pic, &pic);
+        pic.stats = &stats;
+        config_.quality = min_quality;
+        WebPEncode(&config_, &pic);
+        int low_psnr = pic.stats->PSNR[3];
+        WebPPictureFree(&pic);
```
miwucs: This can also be factored out to a sub fuction

src/thumbnailer.cc
```diff
@@ -150,4 +146,119 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+            });
+
+  int final_psnr = -1;
+  WebPData new_webp_data;
+  WebPDataInit(&new_webp_data);
+  WebPPicture pic;
+  WebPPictureInit(&pic);
+  WebPAuxStats stats;
+
+  for (int target_psnr = 99; target_psnr > 0; --target_psnr) {
```
miwucs: This is indeed quite brute force, but it works for a first version!

src/thumbnailer.cc
```diff
@@ -150,4 +146,119 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+        config_.quality = max_quality;
+        WebPEncode(&config_, &pic);
+        int high_psnr = pic.stats->PSNR[3];  // PSNR-all.
+        WebPPictureFree(&pic);
+
+        // Target PSNR is out of range.
+        if (target_psnr > high_psnr || target_psnr < low_psnr) {
+          ok = false;
+          break;
+        }
```
miwucs: If the target_psnr is outside of the range, I think we should fall back to using the closest value anyway (so use min_quality or max_quality) instead of returning an error

src/thumbnailer.cc
```diff
@@ -150,4 +146,119 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+        int max_quality = 100;
+        int final_quality = -1;
+
+        // Get lowest PSNR value possible.
+        WebPPictureInit(&pic);
+        WebPPictureCopy(&frame.pic, &pic);
+        pic.stats = &stats;
+        config_.quality = min_quality;
+        WebPEncode(&config_, &pic);
+        int low_psnr = pic.stats->PSNR[3];
```
miwucs: the PSNR is actually a float

test/thumbnailer_test.cc
```diff
@@ -186,6 +187,48 @@ TEST(ThumbnailerTest, NoisyImageTransparent) {
...
+  auto test_pics = WebPSamplesGenerator(pic_count, transparency, use_randomized)
+                       .GeneratePics();
+
+  for (int i = 0; i < pic_count; ++i) {
+    thumbnailer.AddFrame(*test_pics[i].get(), i * 500);
+  }
+  std::unique_ptr<WebPData, void (*)(WebPData*)> webp_data(new WebPData,
+                                                           WebPDataDelete);
+  WebPDataInit(webp_data.get());
+  thumbnailer.GenerateAnimationEqualPSNR(webp_data.get());
```
miwucs: It would be nice to check that the psnrs are actually more or less equal. But I haven't checked if there's a public function to easily compute the PSNR. @y-guyon can probably help.
Also there's a lot of repetition in all the tests. I think some repetition in tests is generally ok (it's often more readable) but this seems like a lot. Maybe you can factor out some of the code.

y-guyon: There are WebPPlaneDistortion() and WebPPictureDistortion() at your disposal in [encode.h](https://github.com/webmproject/libwebp/blob/master/src/webp/encode.h#L401).
You can either use the latter if you have a WebPPicture nearby, or call the former for each channel otherwise.

ddacw: ~I think it might be `WebPPictureDistortion`.~ nevermind :)

src/thumbnailer.cc
```diff
@@ -51,6 +50,13 @@ Thumbnailer::Status Thumbnailer::AddFrame(const WebPPicture& pic,
   return kOk;
 }
 
+void Thumbnailer::SortFrames() {
+  std::sort(frames_.begin(), frames_.end(),
+            [](const FrameData& A, const FrameData& B) -> bool {
```
miwucs: variables a and b should be in lowercase

src/thumbnailer.h
```diff
@@ -61,8 +63,13 @@ class Thumbnailer {
   // random memory somewhere because the pointer is undefined).
   Status GenerateAnimation(WebPData* const webp_data);
 
+  // Returns the PSNR value of a re-encoded WebPPicture at given quality.
+  int GetPSNR(WebPPicture* const pic, int quality);
+
   // Generates the animation so that all frames have similar PSNR (all) values.
```
miwucs, line 69: It seems you expect GenerateAnimation to be called by the user of the Thumbnailer class before calling GenerateAnimationEqualPSNR

src/thumbnailer.h
```diff
@@ -52,6 +52,8 @@ class Thumbnailer {
   // outlive the last GenerateAnimation() call.
   Status AddFrame(const WebPPicture& pic, int timestamp_ms);
 
+  void SortFrames();
```
miwucs: Please document all public methods, including what they do, and, if relevant, when they should be called (like right now, you expect the caller to call SortFrames() then GenerateAnimation() then GenerateAnimationEqualPSNR(), although I would prefer if that was not the case)

src/thumbnailer.cc
```diff
@@ -141,11 +110,162 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
     }
   }
 
+  config_.quality = final_quality;
+
   if (final_quality == -1) {
```
y-guyon: No need to but if you like to shorten the code you can use
`return (final_quality == -1) ? kByteBudgetError : kOk;`

src/thumbnailer.cc
```diff
@@ -141,11 +110,162 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+
+  WebPData new_webp_data;
+  WebPDataInit(&new_webp_data);
+
+  int high_psnr = -1;
+  int low_psnr = -1;
+  int final_psnr = -1;
+
+  // Find PSNR search range.
+  for (auto& frame : frames_) {
```
y-guyon, line 153: For information I personally dislike the use of `auto` when the iterated type is simple enough (`FrameData` here) but please do as you like, there is no hard rule about it. My reasons are `auto` makes it harder to understand the code in a glimpse and to analyze the code for IDE tools.

src/thumbnailer.cc
```diff
@@ -257,7 +235,38 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
...
+
+Thumbnailer::Status Thumbnailer::SetLoopCount(WebPData* const webp_data) {
+  WebPMuxError err;
+  WebPMuxAnimParams new_params;
+  uint32_t features;
+
+  std::unique_ptr<WebPMux, void (*)(WebPMux*)> mux(WebPMuxCreate(webp_data, 1),
+                                                   WebPMuxDelete);
+
+  if (mux.get() == nullptr) return kMemoryError;
```
miwucs: I don't think you need .get() here (it's done automatically by unique_ptr's == operator)

src/thumbnailer.cc
```diff
@@ -257,7 +235,38 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
...
+  std::cerr << "Final PSNR: " << final_psnr << std::endl;
+
+  if (loop_count_ == 0) return kOk;
+  return SetLoopCount(webp_data);
+}
+
+Thumbnailer::Status Thumbnailer::SetLoopCount(WebPData* const webp_data) {
+  WebPMuxError err;
+  WebPMuxAnimParams new_params;
+  uint32_t features;
```
miwucs: In general we prefer to declare variables close to where they are used, rather than at the beginning of the function (it used to be a thing because it was mandatory in old versions of C, but that's no longer the case).

src/thumbnailer.cc
```diff
@@ -141,11 +110,162 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+  }
+
+  for (int target_psnr = high_psnr; target_psnr >= low_psnr; --target_psnr) {
+    std::cerr << "Target PSNR: " << target_psnr << ". ";
+    bool ok = true;
+
+    WebPAnimEncoderDelete(enc_);
+    enc_ = WebPAnimEncoderNew((frames_[0].pic).width, (frames_[0].pic).height,
+                              &anim_config_);
+    if (enc_ == nullptr) ok = 0;
```
y-guyon: This is too big of an error (memory allocation failure) to just skip this `target_psnr` and carry on, an error status should immediately be returned

src/thumbnailer.cc
```diff
@@ -141,11 +110,162 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+      low_psnr = frame_psnr;
+    }
+  }
+
+  for (int target_psnr = high_psnr; target_psnr >= low_psnr; --target_psnr) {
+    std::cerr << "Target PSNR: " << target_psnr << ". ";
+    bool ok = true;
+
+    WebPAnimEncoderDelete(enc_);
+    enc_ = WebPAnimEncoderNew((frames_[0].pic).width, (frames_[0].pic).height,
```
y-guyon: Why the parenthesis around `frames_[0].pic`?

src/thumbnailer.cc
```diff
@@ -257,7 +235,38 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
...
+  uint32_t features;
+
+  std::unique_ptr<WebPMux, void (*)(WebPMux*)> mux(WebPMuxCreate(webp_data, 1),
+                                                   WebPMuxDelete);
+
+  if (mux.get() == nullptr) return kMemoryError;
+
+  if (WebPMuxGetFeatures(mux.get(), &features) != WEBP_MUX_OK) {
+    return kMemoryError;
+  }
```
miwucs: features is never used so I don't think this is needed

src/thumbnailer.cc
```diff
@@ -257,7 +235,38 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
...
+
+  if (WebPMuxGetAnimationParams(mux.get(), &new_params) != WEBP_MUX_OK) {
+    return kMemoryError;
```
miwucs, line 256: It would be better to return the same error that WebPMuxGetAnimationParams gave. You could define a macro, like (I adapted it from one we have in webp2):
```
#define CHECK_MUX_STATUS(status)              \
  do {                                        \
    const WebPMuxError __S__ = (status);         \
    if (__S__ != WEBP_MUX_OK) {             \
      return __S__;                           \
    }                                         \
  } while (0)
```

Then you can just do:
CHECK_MUX_STATUS(WebPMuxGetAnimationParams(mux.get(), &new_params));

src/thumbnailer.cc
```diff
@@ -141,11 +110,162 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+
+          if (current_psnr <= target_psnr) {
+            frame_final_quality = frame_mid_quality;
+            frame_min_quality = frame_mid_quality + 1;
+          } else {
+            frame_max_quality = frame_mid_quality - 1;
+          }
+        }
+
+        if (frame_final_quality == -1) {
```
y-guyon: What is the meaning of this check? If `frame_lowest_psnr <= target_psnr <= frame_highest_psnr` as verified above, there should be at least one valid quality, right?

src/thumbnailer.cc
```diff
@@ -137,31 +110,55 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
     }
   }
 
+  config_.quality = final_quality;
```
miwucs, line 115: I'm not a fan of the way config_ is used to pass around information between methods, but this could be fixed later.

src/thumbnailer.cc
```diff
@@ -137,31 +110,55 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+  WebPPicture new_pic;
+  WebPPictureInit(&new_pic);
+  WebPPictureCopy(pic, &new_pic);
+
+  WebPAuxStats stats;
+  new_pic.stats = &stats;
+
+  config_.quality = quality;
+  WebPEncode(&config_, &new_pic);
+  int result_psnr = new_pic.stats->PSNR[3];  // PSNR-all.
```
miwucs: std::round(new_pic.stats->PSNR[3])
otherwise you are casting to an int, which is the same as std::floor(new_pic.stats->PSNR[3)

src/thumbnailer.cc
```diff
@@ -137,31 +110,55 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+int Thumbnailer::GetPSNR(WebPPicture* const pic, int quality) {
+  WebPPicture new_pic;
+  WebPPictureInit(&new_pic);
+  WebPPictureCopy(pic, &new_pic);
+
+  WebPAuxStats stats;
+  new_pic.stats = &stats;
+
+  config_.quality = quality;
+  WebPEncode(&config_, &new_pic);
```
miwucs: you should check the return value (it returns false in case of error)

src/thumbnailer.cc
```diff
@@ -141,11 +110,162 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+    if (high_psnr == -1 || frame_psnr > high_psnr) {
+      high_psnr = frame_psnr;
+    }
+    if (low_psnr == -1 || frame_psnr < low_psnr) {
+      low_psnr = frame_psnr;
+    }
+  }
+
+  for (int target_psnr = high_psnr; target_psnr >= low_psnr; --target_psnr) {
+    std::cerr << "Target PSNR: " << target_psnr << ". ";
```
y-guyon: Is this debugging? If it is not an error, `std::cout` might be a better fit

src/thumbnailer.cc
```diff
@@ -141,11 +110,162 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+    bool ok = true;
+
+    WebPAnimEncoderDelete(enc_);
+    enc_ = WebPAnimEncoderNew((frames_[0].pic).width, (frames_[0].pic).height,
+                              &anim_config_);
+    if (enc_ == nullptr) ok = 0;
+
+    // For each frame, find the quality value that produces WebPPicture
+    // having PSNR close to target_psnr.
+    if (ok)
```
y-guyon: Please use { } brackets for such a long block

src/thumbnailer.cc
miwucs: It looks like you haven't addressed this comment. If you disagree (which is fine!) or if you want to fix this in another pull request, please explain that in a comment.

src/thumbnailer.cc
```diff
@@ -137,31 +110,55 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
+
 Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
     WebPData* const webp_data) {
-  // Rearrange frames.
-  std::sort(frames_.begin(), frames_.end(),
-            [](const FrameData& A, const FrameData& B) -> bool {
-              return A.timestamp_ms < B.timestamp_ms;
-            });
+  // GenerateAnimation failed.
+  if (config_.quality == -1) return kByteBudgetError;
```
miwucs: You assume that GenerateAnimation() has been called before GenerateAnimationEqualPSNR(), which is a confusing API. Can't you just call GenerateAnimation() yourself here?
Same thing for SortFrames() which could be called automatically.

test/thumbnailer_test.cc
```diff
@@ -32,47 +32,42 @@ void WebPDataDelete(WebPData* webp_data) {
...
     for (int i = 0; i < pic_count_; ++i) {
-      pics.emplace_back(new WebPPicture, WebPPictureDelete);
-      auto& pic = pics.back();
-      WebPPictureInit(pic.get());
-      pic->use_argb = 1;
-      pic->width = width_;
-      pic->height = height_;
-      WebPPictureImportRGBA(pic.get(), GenerateRGBA(i).data(), width_ * 4);
+      if (thumbnailer.AddFrame(*test_pics[i].get(), i * 500) !=
+          thumbnailer.kOk) {
```
miwucs: Thumbnailer::kOk is better than thumbnailer.kOk because then it's clear that it's a constant (same thing below).

test/thumbnailer_test.cc
```diff
@@ -101,130 +96,117 @@ class WebPSamplesGenerator {
...
-    thumbnailer.AddFrame(*test_pics[i].get(), i * 500);
-  }
   std::unique_ptr<WebPData, void (*)(WebPData*)> webp_data(new WebPData,
                                                            WebPDataDelete);
   WebPDataInit(webp_data.get());
-  thumbnailer.GenerateAnimation(webp_data.get());
+  bool ok = WebPTestGenerator(pic_count, transparency, use_randomized)
+                .GenerateWebPData(webp_data.get());
 
+  EXPECT_EQ(ok, true);
```
miwucs: ASSERT_TRUE(WebPTestGenerator(pic_count, transparency, use_randomized)
                .GenerateWebPData(webp_data.get()));

test/thumbnailer_test.cc
```diff
@@ -31,47 +32,42 @@ void WebPDataDelete(WebPData* webp_data) {
...
+    auto test_pics = GeneratePics();
     for (int i = 0; i < pic_count_; ++i) {
-      pics.emplace_back(new WebPPicture, WebPPictureDelete);
-      auto& pic = pics.back();
-      WebPPictureInit(pic.get());
-      pic->use_argb = 1;
-      pic->width = width_;
-      pic->height = height_;
-      WebPPictureImportRGBA(pic.get(), GenerateRGBA(i).data(), width_ * 4);
+      if (thumbnailer.AddFrame(*test_pics[i].get(), i * 500) !=
```
y-guyon: For `unique_ptr` the `*` operator is overridden so that you do not need the `.get()`

test/thumbnailer_test.cc
```diff
@@ -100,88 +96,117 @@ class WebPSamplesGenerator {
     }
     return rgba;
   }
+
+  // Returns vector of WebPPicture(s).
+  // randomized == true: picture is random noise
+  // randomized == false: picture is random solid color
```
y-guyon: Is `randomized` an argument? I do not see it

test/thumbnailer_test.cc
```diff
@@ -100,88 +96,117 @@ class WebPSamplesGenerator {
...
+  WebPDataInit(webp_data.get());
+  bool ok = WebPTestGenerator(pic_count, transparency, use_randomized)
+                .GenerateWebPData(webp_data.get());
+
+  EXPECT_EQ(ok, true);
+  EXPECT_LE(webp_data->size, kDefaultBudget);
+  EXPECT_GT(webp_data->size, 0);
+}
 
+TEST(ThumbnailerPSNRTest, NoisyImageTransparent) {
```
y-guyon, line 196: To improve further, you might want to take a look at [parameterized GoogleTest cases](http://sandordargo.com/blog/2019/04/24/parameterized-testing-with-gtest) because the only thing that seems to change between the test cases here are `pic_count`, `transparency` and `use_randomized`. The article also gives alternatives (fixtures, loops...)

src/thumbnailer.cc
y-guyon: `WebPPictureCopy()` might also return an error when it fails to allocate more memory; it is not as important as `WebPEncode()` but ideally the returned value should be checked

src/thumbnailer.cc
y-guyon: But `WebPMuxGetAnimationParams()` returns a `WebPMuxError`, do you suggest changing the returned type (currently is `Thumbnailer::Status`)? Maybe the macro needs a `WebPMuxError` to `Thumbnailer::Status` conversion instead

src/thumbnailer.cc
ddacw: If I used `min_quality` or `max_quality` instead of quitting, `thumbnailer` might successfully generate an animation that is not satisfactory (psnr values not approximately close to each other). Anyway, this is now a rare case since I reduced the search size.

src/thumbnailer.cc
ddacw: This was because we could not directly iterate `frames_` with the current struct definition. I'll look into that.

src/thumbnailer.cc
ddacw: I copied some code above w/o noticing it. This was probably overlooked weeks ago.

src/thumbnailer.cc
```diff
@@ -102,43 +101,46 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
     if (new_webp_data.size <= byte_budget_) {
       final_quality = middle;
       WebPDataClear(webp_data);
-      *webp_data = new_webp_data;
+      WebPDataCopy(&new_webp_data, webp_data);
```
miwucs: as Yannis pointed out, this can return an error which should be checked

test/thumbnailer_test.cc
```diff
@@ -120,13 +125,12 @@ TEST(ThumbnailerTest, BlankImageSolid) {
   const int pic_count = 10;
   const uint8_t transparency = 0xff;
   const bool use_randomized = false;
+  const bool equal_PSNR = false;
```
miwucs: should be equal_psnr all in lowercase

test/thumbnailer_test.cc
```diff
@@ -58,16 +58,21 @@ class WebPTestGenerator {
...
-      if (thumbnailer.AddFrame(*test_pics[i].get(), i * 500) !=
-          thumbnailer.kOk) {
+      if (thumbnailer.AddFrame(*test_pics[i], i * 500) != thumbnailer.kOk) {
         return false;
       }
     }
-    return thumbnailer.GenerateAnimation(webp_data) == thumbnailer.kOk;
+    if (!equal_PSNR) {
+      return thumbnailer.GenerateAnimation(webp_data) ==
+             thumbnailer.Status::kOk;
```
miwucs: Thumbnailer::Status::kOk

src/thumbnailer.cc
miwucs: Yeah sorry I didn't realize we were using different error types. But I was just thinking that always returning kMemoryError even if we get an unrelated WebPMuxError (e.g. WEBP_MUX_INVALID_ARGUMENT, WEBP_MUX_BAD_DATA) is not great. So yeah a function/macro to convert between WebPMuxError and Thumbnailer::Status would be good, but probably in another PR.

src/thumbnailer.cc
```diff
@@ -159,102 +161,93 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
...
-        while (frame_min_quality + quality_tolerance <= frame_max_quality) {
-          int frame_mid_quality = (frame_min_quality + frame_max_quality) / 2;
-          int current_psnr = GetPSNR(&frame.pic, frame_mid_quality);
+    for (auto& frame : frames_) {
+      int frame_min_quality = 0;
+      int frame_max_quality = 100;
+      int frame_final_quality = -1;
+
+      int frame_lowest_psnr = GetPSNR(&frame.pic, frame_min_quality);
+      int frame_highest_psnr = GetPSNR(&frame.pic, frame_max_quality);
```
miwucs: const

src/thumbnailer.cc
```diff
@@ -0,0 +1,273 @@
...
+    WebPData new_webp_data;
+    WebPDataInit(&new_webp_data);
+    if (!WebPAnimEncoderAssemble(enc_, &new_webp_data)) {
+      WebPDataClear(&new_webp_data);
+      return kMemoryError;
+    }
+
+    if (all_frames_iterated && new_webp_data.size <= byte_budget_) {
+      final_psnr = target_psnr;
+      WebPDataClear(webp_data);
```
ddacw: @y-guyon I was talking about this part.

test/thumbnailer_test.cc
ddacw: Tests will be left for the next PR.

src/thumbnailer.cc
y-guyon:
```c++
    if (all_frames_iterated && new_webp_data.size <= byte_budget_) {
      // A solution was found, save it.
      final_psnr = target_psnr;
      // Free the memory of the previous solution given by GenerateAnimation().
      WebPDataClear(webp_data);
      // Make 'webp_data' point to the memory location of the new solution.
      *webp_data = new_webp_data;  // The bytes are not copied.
      // No need to free 'new_webp_data' because 'webp_data' takes ownership.
      break;
    } else {
      // Discard this invalid solution and free the associated encoded bitstream bytes.
      WebPDataClear(&new_webp_data);
      std::cerr << std::endl;
    }
```
Is it clearer? Basically discard all invalid solutions until a valid one occurs, then discard the old one and just keep the memory location of the new one. `webp_data` will be freed later by the calling site of `GenerateAnimationEqualPSNR()`, whether it contains the old or the new solution. `webp_data` is actually a `WebPData* const` and not a simple `WebPData` just because it is an output argument.

Keep it as is for now if it works. Make this change in another PR.

src/thumbnailer.cc
```diff
@@ -159,102 +165,96 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
...
-        ok && WebPAnimEncoderAdd(enc_, NULL, frames_.back().timestamp_ms, NULL);
-    ok = ok && WebPAnimEncoderAssemble(enc_, &new_webp_data);
+    if (!WebPAnimEncoderAdd(enc_, NULL, frames_.back().timestamp_ms, NULL)) {
+      return kMemoryError;
+    }
 
-    if (ok && new_webp_data.size <= byte_budget_) {
+    WebPData new_webp_data;
+    WebPDataInit(&new_webp_data);
+    if (!WebPAnimEncoderAssemble(enc_, &new_webp_data)) {
```
miwucs: I guess this can be skipped if all_frames_iterated is false

miwucs (approved):

src/thumbnailer.cc
y-guyon: Was there a memory leak or error? I do not understand why a WebPDataCopy() is needed here. Please revert it in another PR if it works as is.

y-guyon (approved):

miwucs (approved):

PR #18, opened 2020-08-14
Refactor thumbnailer.cc

uyennguyen0903:


src/thumbnailer.cc
```diff
@@ -60,9 +59,15 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationNoBudget(
 
   // Fill the animation.
   for (auto& frame : frames_) {
-    if (!WebPAnimEncoderAdd(enc_, &frame.pic, frame.timestamp_ms, &config_)) {
+    WebPPicture new_pic;
```
y-guyon, line 64: I do not see why a copy is done here neither at line 225.

src/thumbnailer.cc
```diff
@@ -94,29 +99,34 @@ Thumbnailer::Status Thumbnailer::GenerateAnimation(WebPData* const webp_data) {
...
-  config_.quality = final_quality;
+  for (auto& frame : frames_) {
+    frame.config.quality = final_quality;
+  }
+
   return (final_quality == -1) ? kByteBudgetError : kOk;
 }
 
-int Thumbnailer::GetPSNR(WebPPicture* const pic, int quality) {
+int Thumbnailer::GetPSNR(WebPPicture* const pic, WebPConfig config) {
```
y-guyon: `WebPConfig` is not that big but it might be better to use a reference argument here: `const WebPConfig& config`.
`WebPConfig config` would create a copy every time the `GetPSNR()` function is called (unless the compiler optimizes it out). Using a const reference avoid that but checks that it is not modified within the function (safer).

Summary for arguments:
- Use a pass-by-value argument when the type is trivial (`int some_int`, `Thumbnailer::Status some_enum`).
- Use a constant pass-by-reference argument when the type is not (`const WebPConfig& some_conf`).
- Use a constant pointer to a modifiable value when it is an output (`int* const number_of_stuff`, `WebPConfig* const config_to_set`).

src/thumbnailer.cc
```diff
@@ -171,16 +181,17 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
...
     // having PSNR close to target_psnr.
     for (auto& frame : frames_) {
       int frame_min_quality = 0;
       int frame_max_quality = 100;
       int frame_final_quality = -1;
 
-      const int frame_lowest_psnr = GetPSNR(&frame.pic, frame_min_quality);
-      const int frame_highest_psnr = GetPSNR(&frame.pic, frame_max_quality);
+      frame.config.quality = 0;
+      const int frame_lowest_psnr = GetPSNR(&frame.pic, frame.config);
```
y-guyon, line 192: Since `FrameData` contains a `WebPConfig`, `GetPSNR()` could have a single `FrameData` argument now. Up to you.

y-guyon (changes requested):

src/thumbnailer.cc
uyennguyen0903: Since I want to save the original frames for calling `WebPPictureDistortion` later. I think if I used `WebPAnimEncoderAdd(enc_, &frame.pic, frame.timestamp_ms, &frame.config)` the contents of `frame.pic` would be changed.

src/main.cc
```diff
@@ -47,7 +47,7 @@ int main(int argc, char* argv[]) {
   while (input_list >> filename_str >> timestamp_ms) {
     pics.emplace_back(new WebPPicture, WebPPictureFree);
     WebPPictureInit(pics.back().get());
-
+    pics.back().get()->use_argb = 1;
```
miwucs: Put pics.back().get() in a variable since it's used many times

src/thumbnailer.cc
miwucs: Can you add a comment to explain this?

src/thumbnailer.cc
```diff
@@ -60,9 +59,15 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationNoBudget(
 
   // Fill the animation.
   for (auto& frame : frames_) {
-    if (!WebPAnimEncoderAdd(enc_, &frame.pic, frame.timestamp_ms, &config_)) {
+    WebPPicture new_pic;
+    WebPPictureInit(&new_pic);
```
miwucs: According to the documentation of WebPPictureCopy "The 'dst' picture need not be initialized as its content is overwritten."
So I think the call to WebPPictureInit is unnecessary (here and in other places in this file)

miwucs (approved):

src/thumbnailer.cc
y-guyon: From the doc of `WebPAnimEncoderAdd()`:
`frame - (in/out) frame data in ARGB or YUV(A) format. If it is in YUV(A) format, it will be converted to ARGB, which incurs a small loss.`
So if `frame.pic.use_argb` is true, the pixels should not be modified and the copy unnecessary. However it might be safer to copy it anyway and keep the optimization for later; then a comment as Maryla suggested would be appreciated. I agree that the libwebp API appears confusing and inconsistent, it is hard to know what is modified and when.

src/thumbnailer.cc
```diff
@@ -205,13 +217,24 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
...
+
+      if (!WebPAnimEncoderAdd(enc_, &new_pic, frame.timestamp_ms,
+                              &frame.config)) {
         return kMemoryError;
       }
 
-      std::cerr << config_.quality << ' ';
+      WebPPictureFree(&new_pic);
+
+      std::cerr << frame.config.quality << ' ';
```
ddacw, line 236: At the moment, the printed `frame.config.quality` is for debugging. I suggest that we save a list of final quality for each frame, which will be used by our plotting tools. Thoughts? @y-guyon @miwucs

y-guyon: It seems reasonable, it would be stored as a `float quality;` in `FrameData`? Do you plan to output this list through standard output and/or save it to a file?

src/main.cc
```diff
@@ -43,13 +43,15 @@ int main(int argc, char* argv[]) {
   std::ifstream input_list(argv[1]);
   std::string filename_str;
   int timestamp_ms;
+  WebPPicture* current_frame;
```
y-guyon: It can be declared directly at line 50

src/thumbnailer.cc
uyennguyen0903: Is it ok if I store final quality for each frame as `int final_quality` in FrameData ?

src/main.cc
```diff
@@ -64,6 +66,7 @@ int main(int argc, char* argv[]) {
   }
 
   ImgIoUtilWriteFile(output, webp_data.bytes, webp_data.size);
+  WebPPictureFree(current_frame);
```
y-guyon: Each `WebPPicture` in `pics` will be freed, no need to free this `current_frame` (which is just an alias of the last element).

src/thumbnailer.cc
y-guyon: Yes

y-guyon (approved):

PR #19, opened 2020-08-17
Try near-lossless pre-processing to encode each frame.  

uyennguyen0903:


src/main.cc
```diff
@@ -58,12 +60,29 @@ int main(int argc, char* argv[]) {
   WebPData webp_data;
   WebPDataInit(&webp_data);
 
-  if (argc >= 4 && !strcmp(argv[3], "psnr")) {
+  bool try_equal_psnr = false;
+  bool try_near_lossless = false;
+
+  if (argc >= 4) {
```
miwucs: this if is not really needed

src/thumbnailer.cc
```diff
@@ -39,16 +39,81 @@ Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc_); }
...
                                           int timestamp_ms) {
   // Verify dimension of frames.
-  if (!frames_.empty() && (pic.width != (frames_[0].pic).width ||
-                           pic.height != (frames_[0].pic).height)) {
+  if (!frames_.empty() && (pic.width != frames_[0].pic.width ||
+                           pic.height != frames_[0].pic.height)) {
     return kImageFormatError;
   }
   WebPConfig new_config;
   WebPConfigInit(&new_config);
```
skal65535: this call can fail in case of library version mismatch. An assert could be useful.

src/thumbnailer.cc
```diff
@@ -39,16 +39,81 @@ Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc_); }
...
+  new_config.show_compressed = 1;
   frames_.push_back({pic, timestamp_ms, new_config});
   return kOk;
 }
 
+int Thumbnailer::GetSize(WebPPicture* const pic, const WebPConfig& config) {
+  const int failure = -1;
+
+  WebPPicture new_pic;
+  WebPPictureInit(&new_pic);
```
skal65535: same here: this call can (rarely) fail. Could be incorporated with the if(...) below.

src/thumbnailer.cc
```diff
@@ -39,16 +39,81 @@ Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc_); }
...
   }
   WebPConfig new_config;
   WebPConfigInit(&new_config);
+  new_config.show_compressed = 1;
   frames_.push_back({pic, timestamp_ms, new_config});
   return kOk;
 }
 
+int Thumbnailer::GetSize(WebPPicture* const pic, const WebPConfig& config) {
+  const int failure = -1;
```
skal65535: why 'const'? if it's a real constant, use 'kFailure'

src/thumbnailer.cc
```diff
@@ -39,16 +39,81 @@ Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc_); }
...
+  }
+
+  WebPAuxStats stats;
+  new_pic.stats = &stats;
+  if (!WebPEncode(&config, &new_pic)) {
+    WebPPictureFree(&new_pic);
+    return failure;
+  }
+
+  int frame_size = new_pic.stats->coded_size;
```
skal65535: const int frame_size ...

src/thumbnailer.cc
skal65535: actually, you don't need this call to WebPPictureInit() if you're calling WebPPictureCopy() directly

src/thumbnailer.cc
```diff
@@ -39,16 +39,81 @@ Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc_); }
...
+  const int failure = -1;
+  WebPPicture new_pic;
+
+  if (!WebPPictureCopy(pic, &new_pic) || (!WebPEncode(&config, &new_pic))) {
+    WebPPictureFree(&new_pic);
+    return failure;
+  }
+
+  float distortion_result[5];
+  int result_psnr;
```
skal65535: should be a 'float'

src/thumbnailer.cc
```diff
@@ -39,16 +39,80 @@ Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc_); }
...
   // Verify dimension of frames.
-  if (!frames_.empty() && (pic.width != (frames_[0].pic).width ||
-                           pic.height != (frames_[0].pic).height)) {
+  if (!frames_.empty() && (pic.width != frames_[0].pic.width ||
+                           pic.height != frames_[0].pic.height)) {
     return kImageFormatError;
   }
   WebPConfig new_config;
-  WebPConfigInit(&new_config);
+  assert(WebPConfigInit(&new_config));
```
y-guyon: You should not put any behavior inside assertions because it will be executed only in debug.
Replace with `if (!WebPConfigInit(&new_config)) assert(false);` or similar.

src/thumbnailer.cc
```diff
@@ -39,16 +39,80 @@ Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc_); }
...
+  new_config.show_compressed = 1;
   frames_.push_back({pic, timestamp_ms, new_config});
   return kOk;
 }
 
+int Thumbnailer::GetSize(WebPPicture* const pic, const WebPConfig& config) {
+  int failure = -1;
```
y-guyon: I think @skal65535 was referring to the style guide : https://google.github.io/styleguide/cppguide.html#Constant_Names

So either `int failure = -1;` or `const int kFailure = -1;` is expected, but since the value is indeed a constant (could be an `enum`) the latter might be more appropriate. You can even declare it only once with a `static` keyword at the beginning of this file and use it everywhere, to avoid declaring it in each function.

src/thumbnailer.cc
```diff
@@ -39,16 +39,80 @@ Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc_); }
...
+  WebPPictureFree(&new_pic);
+
+  return frame_size;
+}
+
+float Thumbnailer::GetPSNR(WebPPicture* const pic, const WebPConfig& config) {
+  int failure = -1;
+  WebPPicture new_pic;
+
+  if (!WebPPictureCopy(pic, &new_pic) || (!WebPEncode(&config, &new_pic))) {
```
y-guyon: Unnecessary parenthesis

src/thumbnailer.cc
```diff
@@ -269,25 +314,91 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
...
+    // Only try binary search if near-lossles encoding with pre-processing = 0
+    // is feasible in order to save execution time.
+    if (anim_size - curr_size + new_size <= byte_budget_) {
+      // Binary search for near-lossless's pre-processing value.
+      while (min_near_ll <= max_near_ll) {
+        int mid_near_ll = (min_near_ll + max_near_ll) / 2;
+        frame.config.near_lossless = mid_near_ll;
+        new_size = GetSize(&frame.pic, frame.config);
+        if (anim_size - curr_size + new_size <= byte_budget_) {
+          float new_psnr = GetPSNR(&frame.pic, frame.config);
```
miwucs: const

src/thumbnailer.h
```diff
@@ -64,17 +65,29 @@ class Thumbnailer {
...
+  // Returns -1 on errors.
+  int GetSize(WebPPicture* const pic, const WebPConfig& config);
 
   // Generates the animation so that all frames have similar PSNR (all) values.
   // In case of failure, takes the animation generated by GenerateAnimation() as
   // result.
   Status GenerateAnimationEqualPSNR(WebPData* const webp_data);
 
+  // Tries near-lossless to encode each frame. In case of failure, keeps the
+  // latest lossy encoded frame.
```
miwucs: You should add that either GenerateAnimation or GenerateAnimationEqualPSNR must be called before this (it's important to document how the API must be used).

src/thumbnailer.h
```diff
@@ -64,17 +65,29 @@ class Thumbnailer {
...
   Status SetLoopCount(WebPData* const webp_data);
 
-  // List of final quality for each frames.
-  std::vector<int> final_qualities;
+  struct FrameResult {
+    int size;
+    int quality;
+  };
+  // Vector of frames's final results.
+  std::vector<FrameResult> results_;
```
miwucs: does this have to be public?

src/thumbnailer.cc
```diff
@@ -39,16 +39,80 @@ Thumbnailer::~Thumbnailer() { WebPAnimEncoderDelete(enc_); }
...
+  if (!WebPPictureDistortion(pic, &new_pic, 0, distortion_result)) {
+    result_psnr = failure;
+  } else {
+    result_psnr = distortion_result[4];  // PSNR-all.
+  }
+  WebPPictureFree(&new_pic);
+  return result_psnr;
+}
+
+Thumbnailer::Status Thumbnailer::SetLoopCount(WebPData* const webp_data) {
```
y-guyon, line 94: In the future please avoid moving functions around or refactor them in a separate pull request. Otherwise it is hard to know what needs to be reviewed.

src/thumbnailer.cc
y-guyon: For another pull request: encoding the image a second time just to get the distortion should be avoided.

src/thumbnailer.h
y-guyon: Also is there a reason why this is a separate `vector` and not part of `FrameData`?

uyennguyen0903: the `results_ vector` is probably unnecessary, I will move the `size` and `quality` to `FrameData`.

src/thumbnailer.cc
uyennguyen0903: I'm not sure why we need `const float new_psnr = GetPSNR(&frame.pic, frame.config);` here. Could you explain please ?

miwucs: It's not really required, it's just a question of style. We like to use const on all variables that are only assigned once. The idea is that it makes the code easier to read, because the person reading knows instantly that this is a constant value and that they don't need to track this variable's value in their head. So I think a lot of teams at Google (including us) use const for all variables that never change, even if it's not required by the style guide.

src/thumbnailer.cc
ddacw: This constant (`failure`) will be returned by functions that should return nonnegative integer (size, psnr, etc.). Should we declare an `enum kUnsignedError = -1`?

y-guyon: You may although `enum`s are typically used for several meanings (`Status::kOk`, `Status::kMemoryError` etc.) and not for actual numbers (-1).
I would suggest `static const int kFailure = -1;` or using the already existing `Thumbnailer::Status` everywhere instead (requires adding output arguments to `GetPSNR()`, `GetSize()` because the returned value will now only be the status).

uyennguyen0903:
PTAL.

y-guyon (approved):

src/thumbnailer.cc
uyennguyen0903: For later, how should we get the `PSNR` without encoding the image a second time ?

miwucs (approved):

src/thumbnailer.cc
y-guyon: You could store the encoded bitstream of `GetSize()` and then decode+disto it only if needed.

PR #21, opened 2020-08-18
Add tests for near-lossless and fix thumbnailer.cc

uyennguyen0903:
- Add tests for `TryNearLossless()`.
- Replace animation size in `TryNearLossless()` with the sum of encoded-frame sizes.


src/thumbnailer.cc
```diff
@@ -314,7 +314,10 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
 }
 
 Thumbnailer::Status Thumbnailer::TryNearLossless(WebPData* const webp_data) {
-  int anim_size = webp_data->size;
+  int anim_size = 0;
```
y-guyon, line 319: Add a short comment on why `webp_data->size` is not used instead

y-guyon (approved):

PR #22, opened 2020-08-19
Compare thumbnails.

ddacw:


src/thumbnailer.h
```diff
@@ -84,6 +86,11 @@ class Thumbnailer {
 
   Status SetLoopCount(WebPData* const webp_data);
 
+  // Takes two thumbnails as WebPData. Prints various statistics regarding
+  // the differences in PSNR between two thumbnails.
+  Status CompareThumbnail(WebPData* const webp_data_ref,
```
y-guyon: `CompareThumbnail()`, `AnimData2Pictures()` and `AnimData2PSNR()` should not be part of `Thumbnailer` if they are in a separate library. I advise separate functions in a compare.h, utils.h or similar because they are not really related to `Thumbnailer` (but may be in the same library for convenience).

src/thumbnailer.h
```diff
@@ -84,6 +86,11 @@ class Thumbnailer {
 
   Status SetLoopCount(WebPData* const webp_data);
 
+  // Takes two thumbnails as WebPData. Prints various statistics regarding
```
y-guyon: `CompareThumbnail()` should output the stats rather than print them. It is the job of main.cc (or another binary) to print or aggregate them.

src/thumbnailer_compare.cc
```diff
@@ -0,0 +1,125 @@
...
+      WebPAnimDecoderNew(webp_data, NULL), WebPAnimDecoderDelete);
+  if (dec == NULL) {
+    std::cerr << "Error parsing image." << std::endl;
+    return kMemoryError;
+  }
+
+  WebPAnimInfo anim_info;
+  if (!WebPAnimDecoderGetInfo(dec.get(), &anim_info)) {
+    std::cerr << "Error getting global info about the animation";
+    return kMemoryError;
```
y-guyon: In another PR, rename `kMemoryError` to `kGenericError` or add more `Thumbnailer::Status` because returning a memory error might be misleading here.

src/thumbnailer_compare.cc
```diff
@@ -0,0 +1,125 @@
...
+  // WebPData doesn't contain any frames.
+  if (psnr.empty()) return kOk;
+
+  const int frame_count = psnr.size();
+
+  for (int i = 0; i < frame_count; ++i) {
+    psnr_diff.push_back(psnr[i] - psnr_ref[i]);
+  }
+
+  std::sort(psnr_diff.begin(), psnr_diff.end());
```
y-guyon: Sorting is unnecessary, `std::minmax_element()` is enough and faster

ddacw: I sorted `psnr_diff` to get the median.

y-guyon: Oh right I forgot. You have `std::nth_element()` for that but do not bother indeed, sorting is fine.

src/thumbnailer_utils.cc
```diff
@@ -0,0 +1,193 @@
...
+  std::unique_ptr<WebPAnimDecoder, void (*)(WebPAnimDecoder*)> dec(
+      WebPAnimDecoderNew(webp_data, NULL), WebPAnimDecoderDelete);
+  if (dec == NULL) {
+    std::cerr << "Error parsing image." << std::endl;
+    return kMemoryError;
+  }
+
+  WebPAnimInfo anim_info;
+  if (!WebPAnimDecoderGetInfo(dec.get(), &anim_info)) {
+    std::cerr << "Error getting global info about the animation";
```
y-guyon: Missing `std::endl`

src/thumbnailer_utils.cc
```diff
@@ -0,0 +1,193 @@
...
+UtilsStatus CompareThumbnail(
+    const std::vector<EnclosedWebPPicture>& original_pics,
+    WebPData* const webp_data_1, WebPData* const webp_data_2,
+    ThumbnailDiffPSNR* const diff) {
+  if (original_pics.empty()) {
+    std::cerr << "Thumbnail doesn't contain any frames.";
+    return kOk;
+  }
+
+  if (diff == NULL) return kMemoryError;
```
y-guyon: Use `NULL` only when dealing with C API (`WebPAnimDecoderNew()` etc.) and `nullptr` everywhere else.

src/thumbnailer_utils.h
```diff
@@ -0,0 +1,89 @@
...
+  float min_psnr, max_psnr, mean_psnr, median_psnr;
+};
+
+// Stores the difference in PSNR values between two thumbnails,
+// with various statistics.
+struct ThumbnailDiffPSNR {
+  std::vector<float> psnr_diff;
+  float max_psnr_increase;
+  float max_psnr_decrease;
+  float sum_psnr_diff;
```
y-guyon: Is this useful on top of `mean_psnr_diff`?

y-guyon (approved):

src/thumbnailer_utils.cc
```diff
@@ -21,13 +21,13 @@ UtilsStatus AnimData2Pictures(WebPData* const webp_data,
   std::unique_ptr<WebPAnimDecoder, void (*)(WebPAnimDecoder*)> dec(
       WebPAnimDecoderNew(webp_data, NULL), WebPAnimDecoderDelete);
   if (dec == NULL) {
-    std::cerr << "Error parsing image." << std::endl;
+    std::cerr << "Error parsing image.\n";
```
y-guyon, line 24: It is odd to mix the old C `"\n"` with the newer C++ `std::cout` syntax but I do not mind.

y-guyon (approved):

PR #24, opened 2020-08-24
Implement the slope optimization for lossy to produce animation.

uyennguyen0903:
- Find the median slope on the RD-curve.
- Use the median slope to optimize the binary search with lossy compression mode and save the byte budget for near-lossless.
- Re-encode frames with lossy to get higher PSNR if possible after calling `TryNearLossless()`.


src/main.cc
```diff
@@ -62,23 +62,30 @@ int main(int argc, char* argv[]) {
...
   bool try_equal_psnr = false;
   bool try_near_lossless = false;
+  bool slope_optim = false;
 
   for (int c = 3; c < argc; c++) {
     if (!strcmp(argv[c], "psnr")) {
       try_equal_psnr = true;
     } else if (!strcmp(argv[c], "try_near_lossless")) {
       try_near_lossless = true;
+    } else if (!strcmp(argv[c], "slope_optimization")) {
```
miwucs, line 75: add some documentation at some point

src/main.cc
```diff
@@ -62,23 +62,30 @@ int main(int argc, char* argv[]) {
...
+      slope_optim = true;
     }
   }
 
-  if (try_equal_psnr) {
-    thumbnailer.GenerateAnimationEqualPSNR(&webp_data);
-  } else {
-    thumbnailer.GenerateAnimation(&webp_data);
-  }
+  if (!slope_optim) {
```
miwucs: Minor detail, it's considered good practice to use simpler if statements if possible, like 
if(slope_optim) {
  // stuff 1
} else {
  // stuff 2
}

instead of
if (!slope_optim) {
  // stuff 2
} else {
  // stuff 1
}

src/thumbnailer.h
```diff
@@ -84,14 +87,43 @@ class Thumbnailer {
...
+  // Generates the animation with lossy compression mode and optimizes the
+  // resulting qualities using the median slope in order to save byte budget for
+  // calling TryNearLossless().
+  Status LossyEncodeSlopeOptim(WebPData* const webp_data);
+
+  // Tries to re-encode each frame with the lossy compression mode to find the
+  // better PSNR values if possible. Both LossyEncodeSlopeOptim() and
+  // TryNearLossless() must be respectively called before to generate the
+  // animation.
+  Status LossyEncodeNoSlopeOptim(WebPData* const Webp_data);
```
miwucs: These should be private. Nice documentation by the way!

src/thumbnailer_slope_optim.cc
```diff
@@ -0,0 +1,253 @@
...
+Thumbnailer::Status Thumbnailer::FindMedianSlope(float* const median_slope) {
+  std::vector<float> slopes;
+
+  for (auto& frame : frames_) {
+    frame.config.quality = 100;
+    float psnr_100;  // pic's psnr value with quality = 100.
+    int size_100;    // pic'size with quality = 100.
+    if (GetPictureStats(frame.pic, frame.config, &size_100, &psnr_100) != kOk) {
+      return kStatsError;
+    };
```
miwucs: extra ;

src/thumbnailer_slope_optim.cc
```diff
@@ -0,0 +1,253 @@
...
+    slopes.push_back(pic_final_slope);
+  }
+
+  std::sort(slopes.begin(), slopes.end());
+  *median_slope = slopes[slopes.size() / 2];
+
+  return kOk;
+}
+
+Thumbnailer::Status Thumbnailer::ComputeSlope(const int& ind,
```
miwucs: There is no need to use const references for simple types like ints. We prefer to pass by value in this case (so simply using "int ind" and so on)

src/thumbnailer_slope_optim.cc
```diff
@@ -0,0 +1,253 @@
...
+  // Use binary search with slope optimization to find quality values that makes
+  // the animation fit the given byte budget. The quality value for each frame
+  // can be different.
+  while (min_quality <= max_quality && !optim_list.empty()) {
+    int mid_quality = (min_quality + max_quality) / 2;
+
+    std::queue<int> new_optim_list;
+
+    while (!optim_list.empty()) {
+      int curr_frame = optim_list.front();
```
miwucs: const

src/thumbnailer_slope_optim.cc
```diff
@@ -0,0 +1,253 @@
...
+
+Thumbnailer::Status Thumbnailer::FindMedianSlope(float* const median_slope) {
+  std::vector<float> slopes;
+
+  for (auto& frame : frames_) {
+    frame.config.quality = 100;
+    float psnr_100;  // pic's psnr value with quality = 100.
+    int size_100;    // pic'size with quality = 100.
+    if (GetPictureStats(frame.pic, frame.config, &size_100, &psnr_100) != kOk) {
+      return kStatsError;
```
miwucs: I think in this kind of situation you should propagate the original error you got instead of returning a new one. Otherwise the original error is lost. It might be useful to have a macro as I've mentioned in a previous review (it was a mistake from me at the time since the error types were different, but here the error types are the same).
Something like:
#define CHECK_THUMBNAILER_STATUS(status)              \
  do {                                        \
    const Thumbnailer::Status  __S__ = (status);         \
    if (__S__ != Thumbnailer::kOk) {             \
      return __S__;                           \
    }                                         \
  } while (0)

Then you can just do:
CHECK_THUMBNAILER_STATUS(GetPictureStats(frame.pic, frame.config, &size_100, &psnr_100));

src/thumbnailer_slope_optim.cc
```diff
@@ -0,0 +1,253 @@
...
+  if (FindMedianSlope(&limit_slope) != kOk) {
+    return kSlopeOptimError;
+  }
+
+  int min_quality = minimum_lossy_quality_;
+  int max_quality = 100;
+  WebPData new_webp_data;
+  WebPDataInit(&new_webp_data);
+
+  std::queue<int> optim_list;  // Queue of frames needed to find the quality in
```
miwucs: I think this and new_optim_list can just be vectors that you iterate over instead of popping stuff (unless I missed something)

src/thumbnailer_slope_optim.cc
```diff
@@ -0,0 +1,253 @@
...
+        int curr_frame = new_optim_list.front();
+        new_optim_list.pop();
+        frames_[curr_frame].final_quality = mid_quality;
+        optim_list.push(curr_frame);
+      }
+      WebPDataClear(webp_data);
+      *webp_data = new_webp_data;
+      min_quality = mid_quality + 1;
+    } else {
+      optim_list = new_optim_list;
```
miwucs: I'm not sure I get this. During the next pass, we will try changing the quality of the frames that changed during this pass, right? But if we're lowering the quality, some frames that didn't get changed during the last pass because the slope was too low might need to be updated, no?

src/thumbnailer_slope_optim.cc
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  WebPData new_webp_data;
+  WebPDataInit(&new_webp_data);
+
+  // For each frame, find the best quality value that can produce the higher
+  // PSNR than the current one if possible.
+  for (auto& frame : frames_) {
+    int min_quality = 0;
+    int max_quality = 100;
```
miwucs: Is it really necessary to search over the whole range? Maybe it's enough to search around a window of, say, +/- 10 around the existing quality setting?

src/thumbnailer_slope_optim.cc
```diff
@@ -0,0 +1,253 @@
...
+      float new_psnr;
+      frame.config.quality = mid_quality;
+      if (GetPictureStats(frame.pic, frame.config, &new_size, &new_psnr) !=
+          kOk) {
+        return kStatsError;
+      }
+
+      if (new_psnr > frame.final_psnr || ((new_psnr == frame.final_psnr) &&
+                                          (new_size <= frame.encoded_size))) {
+        if (anim_size - frame.encoded_size + new_size <= byte_budget_) {
```
miwucs: This might tend to use all the available budget for the first frames and leave no budget for the other frames. Maybe we should try to allocate some budget to every frame, like:
int extra_budget = (byte_budget_ - anim_size) / num_remaining_frames;
if (new_size - frame.encoded_size  < extra_budget)

src/thumbnailer.h
```diff
@@ -84,14 +87,43 @@ class Thumbnailer {
 
   Status SetLoopCount(WebPData* const webp_data);
 
+  // Generates the animation with the slope optimization for the RD-curve.
+  // Either lossy and near-lossless compression modes will be used for each
+  // frame.
+  Status GenerateAnimationSlopeOptim(WebPData* const webp_data);
```
miwucs, line 91: Can you add some tests for this method, at least to make sure valgrind catches any memory leaks.

miwucs (commented):
Interesting algorithm. I'm curious to see what the results are.

src/thumbnailer_slope_optim.cc
uyennguyen0903: During the next pass, the quality of the frames that have `final_quality = -1` or `slope(min_quality, max_quality) >= limit_slope`  will be changed. When we're lowering the quality, it means the frames in `new_optim_list` did not get changed, so the `optim_list` in the next pass will be identical to the previous `new_optim_list`. Also, when we're lowering the `max_quality`, the new slope for each frame will be clearly higher than the older one so it will pass the condition at line 146.

src/thumbnailer_slope_optim.cc
uyennguyen0903: You are right, I will change `optim_list` and `new_optim_list` into `vectors`.

src/thumbnailer.h
uyennguyen0903: Yes, I will.

uyennguyen0903:
PTAL.

src/thumbnailer.cc
```diff
@@ -14,6 +14,12 @@
 
 #include "thumbnailer.h"
 
+#define CONVERT_WEBP_MUX_ERROR_STATUS(webp_mux_error) \
```
miwucs: maybe remove "ERROR" to make it a bit shorter

src/thumbnailer.cc
```diff
@@ -318,12 +306,14 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualPSNR(
 }
 
 Thumbnailer::Status Thumbnailer::TryNearLossless(WebPData* const webp_data) {
-  // The webp_data->size is not used here instead since it is higher than the
-  // sum of encoded-frame sizes.
+  // The webp_data->size and the sum of encoded-frame sizes are inconsistent,
+  // therefore consider the bigger one as the current animation size to ensure
+  // the resulting animation size fit the byte budget.
```
miwucs, line 309: good find!

src/thumbnailer_slope_optim.cc
```diff
@@ -18,16 +18,17 @@ namespace libwebp {
...
 
-  if (anim_status == kOk) {
-    anim_status = LossyEncodeNoSlopeOptim(webp_data);
-  }
+  int curr_anim_size = webp_data->size;
+  do {
+    CHECK_THUMBNAILER_STATUS(LossyEncodeNoSlopeOptim(webp_data));
+    if (curr_anim_size == webp_data->size) break;
+    curr_anim_size = webp_data->size;
+  } while (0);
```
miwucs: What is this while for?

src/thumbnailer_slope_optim.cc
```diff
@@ -131,36 +122,28 @@ Thumbnailer::Status Thumbnailer::LossyEncodeSlopeOptim(
...
     int mid_quality = (min_quality + max_quality) / 2;
 
-    std::queue<int> new_optim_list;
-
-    while (!optim_list.empty()) {
-      int curr_frame = optim_list.front();
-      optim_list.pop();
+    std::vector<int> new_optim_list;
 
+    for (int& curr_frame : optim_list) {
```
miwucs: Doesn't need to be a reference (same in new_optim_list loop below)

src/thumbnailer_slope_optim.cc
```diff
@@ -190,33 +171,38 @@ Thumbnailer::Status Thumbnailer::LossyEncodeSlopeOptim(
...
-          kOk) {
-        return kStatsError;
-      }
+      CHECK_THUMBNAILER_STATUS(
+          GetPictureStats(frame.pic, frame.config, &new_size, &new_psnr));
 
       if (new_psnr > frame.final_psnr || ((new_psnr == frame.final_psnr) &&
                                           (new_size <= frame.encoded_size))) {
-        if (anim_size - frame.encoded_size + new_size <= byte_budget_) {
+        float extra_budget = (byte_budget_ - anim_size) / num_remaining_frames;
```
miwucs: const

miwucs: Are we guaranteed that anim_size is actually smaller byte_budget? Otherwise extra_budget could be negative.

src/thumbnailer_slope_optim.cc
```diff
@@ -131,36 +122,28 @@ Thumbnailer::Status Thumbnailer::LossyEncodeSlopeOptim(
...
 
     if (new_webp_data.size <= byte_budget_) {
-      while (!new_optim_list.empty()) {
-        int curr_frame = new_optim_list.front();
-        new_optim_list.pop();
+      optim_list.clear();
+      for (int& curr_frame : new_optim_list) {
         frames_[curr_frame].final_quality = mid_quality;
-        optim_list.push(curr_frame);
+        optim_list.push_back(curr_frame);
```
miwucs: In both branches of the if, this ends up being equivalent to optim_list = new_optim_list, right? It would be clearer to have it written this way

miwucs (approved):

src/thumbnailer_slope_optim.cc
uyennguyen0903: I think that `anim_size` is smaller than `byte_budget` in most cases. Might it be better if I write something like : `if (anim_size > byte_budget) return kOk;` [here](https://github.com/googleinterns/step255-2020/blob/b02b1727d9b0135538ce6c3196721cbe1bb11007/src/thumbnailer_slope_optim.cc#L182) in order to guarantee the extra_budget be positive and do nothing otherwise ?

src/thumbnailer_slope_optim.cc
uyennguyen0903: That's right, I will remove lines 143, 146 and then write `optim_list = new_optim_list;` right below the end of the `if`.

src/thumbnailer_slope_optim.cc
miwucs: Yes that sounds pretty reasonable.

src/thumbnailer_slope_optim.cc
uyennguyen0903: the `while` loop is for guarantee that we cannot get better results with `LossyEncodeNoSlopeOptim` anymore.

miwucs: This is a while(0) so it will only run once. If it was a while(true), you would need some guarantee that it will stop.

uyennguyen0903: It's my mistake, I thought that `while(true)` and `while(0)` were similar :) .

src/thumbnailer_slope_optim.cc
```diff
@@ -26,7 +26,7 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationSlopeOptim(
     CHECK_THUMBNAILER_STATUS(LossyEncodeNoSlopeOptim(webp_data));
     if (curr_anim_size == webp_data->size) break;
     curr_anim_size = webp_data->size;
-  } while (0);
+  } while (true);
```
miwucs: I think it would be safer to use a for loop so that we're sure we don't do too many iterations, and more importantly don't loop forever.
So something like:
const int kMaxIter = 5;
for (int i = 0; i < KMaxIter; ++i) {
  CHECK_THUMBNAILER_STATUS(LossyEncodeNoSlopeOptim(webp_data));
  if (curr_anim_size == webp_data->size) break;	    if (curr_anim_size == webp_data->size) break;
  curr_anim_size = webp_data->size;	    curr_anim_size = webp_data->size;
}

uyennguyen0903: Done!

miwucs (approved):

PR #25, opened 2020-08-25
Improve CompareThumbnail

ddacw:
TODO: create a wrapper for CompareThumbnail.


src/thumbnailer_utils.h
```diff
@@ -45,6 +45,11 @@ enum UtilsStatus {
   kGenericError  // For other errors.
 };
 
+struct Frame {
+  EnclosedWebPPicture pic;
+  int timestamp;
```
y-guyon: Rename the variable or comment whether it is the starting or ending timestamp

src/thumbnailer_utils.h
```diff
@@ -45,6 +45,11 @@ enum UtilsStatus {
   kGenericError  // For other errors.
 };
 
+struct Frame {
+  EnclosedWebPPicture pic;
```
y-guyon, line 55: Maybe `EnclosedWebPPicture` is not needed now that `Frame` exists? Up to you.

src/thumbnailer_utils.cc
```diff
@@ -54,72 +55,73 @@ UtilsStatus AnimData2Pictures(WebPData* const webp_data,
...
+  for (int i = 0; i < frame_count; ++i) {
+    const Frame& original_frame = original_frames[i];
+
+    // Check if the next frame of new_frames matches original_frame,
+    // based on their timestamps. This is because consecutive original frames
+    // having the exact same WebPPicture are merged into one.
+    if (new_frame_index + 1 < new_frames.size() &&
+        new_frames[new_frame_index + 1].timestamp == original_frame.timestamp) {
+      ++new_frame_index;
+    }
```
y-guyon, line 85: You should check or assert than the timestamps match before calling `WebPPictureDistortion()`.

src/thumbnailer_utils.cc
```diff
@@ -54,72 +55,73 @@ UtilsStatus AnimData2Pictures(WebPData* const webp_data,
...
+  if (data2frames_status != kOk) return data2frames_status;
+
+  const int frame_count = original_frames.size();
+  int new_frame_index = 0;
+  for (int i = 0; i < frame_count; ++i) {
+    const Frame& original_frame = original_frames[i];
+
+    // Check if the next frame of new_frames matches original_frame,
+    // based on their timestamps. This is because consecutive original frames
+    // having the exact same WebPPicture are merged into one.
```
y-guyon: I do not think it is worth it to handle the opposite case but it would be safer to mention it here in case it becomes an issue: an `original_frame` so long that it was split into two `new_frames` (log2(duration) > 24 bits as specified at https://developers.google.com/speed/webp/docs/riff_container). I did not check but I guess libwebp does not currently split frames, behavior is probably undefined.

src/utils/thumbnailer_compare.cc
```diff
@@ -0,0 +1,80 @@
...
+  // Generate new thumbnails and compare to the reference thumbnail.
+  int options_count = 1;
+  WebPData webp_data;
+  WebPDataInit(&webp_data);
+  for (int i = 0; i < options_count; ++i) {
+    libwebp::Thumbnailer thumbnailer;
+    for (libwebp::Frame& frame : frames) {
+      thumbnailer.AddFrame(*frame.pic, frame.timestamp);
+    }
+    thumbnailer.GenerateAnimationEqualPSNR(&webp_data);
```
ddacw: I think we should have `Thumbnailer` taking an `int option` and call the corresponding method, so we don't have to specify methods.

src/utils/thumbnailer_compare.cc
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  // Generate reference thumbnail.
+  libwebp::Thumbnailer thumbnailer_ref;
+  for (libwebp::Frame& frame : frames) {
+    thumbnailer_ref.AddFrame(*frame.pic, frame.timestamp);
+  }
+  WebPData webp_data_ref;
+  WebPDataInit(&webp_data_ref);
+  thumbnailer_ref.GenerateAnimation(&webp_data_ref);
```
y-guyon: There are returned `Status` that are not checked in this file.
If you want the compiler to warn you about that, you may use the `[[nodiscard]]` attribute on the `Status` enum declaration or on some specific functions [1] 

[1] https://en.cppreference.com/w/cpp/language/attributes/nodiscard

src/utils/thumbnailer_compare.cc
```diff
@@ -0,0 +1,80 @@
...
+    ReadImage(filename_str.c_str(), pic);
+  }
+  if (frames.empty()) {
+    std::cerr << "No input file(s) for generating animation." << std::endl;
+    return 1;
+  }
+
+  // Generate reference thumbnail.
+  libwebp::Thumbnailer thumbnailer_ref;
+  for (libwebp::Frame& frame : frames) {
```
y-guyon: I suggest the `libwebp::Frame& frame` to be `const` to make sure input data is not modified.

y-guyon (approved):

uyennguyen0903:
I modified `thumbnailer.cc` in my last PR and merged it on branch master. Do not forget to rebase the code to avoid conflicts in `thumbnailer.cc`.

src/utils/thumbnailer_compare.cc
y-guyon: It might be more convenient yes, you could also translate that into an enum so it can be easily cast from an integer given by the command line but still bear more meaning than a raw int in the code.

src/thumbnailer.h
```diff
@@ -57,32 +57,41 @@ class Thumbnailer {
...
     kStatsError,        // In case of error while getting frame's size and PSNR.
     kWebPMuxError,      // In case of error related to WebPMux object.
-    kSlopeOptimError    // In case of error while using slope optimization to
+    kSlopeOptimError,   // In case of error while using slope optimization to
                         // generate animation.
+    kGenericError       // For other errors.
   };
 
+  enum Method { kEqualQuality = 0, kEqualPSNR, kSlopeOptim };
+  static constexpr Method MethodList[] = {kEqualQuality, kEqualPSNR,
```
y-guyon: If this is used only at one place, maybe just inline it there

src/utils/thumbnailer_utils.h
```diff
@@ -45,10 +45,10 @@ typedef std::unique_ptr<WebPPicture, void (*)(WebPPicture*)>
 
 namespace libwebp {
 
-enum UtilsStatus {
-  kOk = 0,       // On success.
-  kMemoryError,  // In case of memory error.
-  kGenericError  // For other errors.
+enum [[nodiscard]] UtilsStatus{
```
y-guyon, line 48: This does not seem to be taken into account if `GenerateAnimationEqualQuality()` can be called without using its return value, I am unsure why, maybe there is a compile flag to enable or something

y-guyon (approved):

src/utils/thumbnailer_utils.h
y-guyon: You are right this is not the main Status enum, my bad.

src/thumbnailer.h
y-guyon: By inlining it I meant something similar to
```c++
for (const libwebp::Thumbnailer::Method& method :
     {kEqualQuality, kEqualPSNR, kSlopeOptim}) {
  ...
}
```
and get rid of `MethodList` but this is just a suggestion. Otherwise call it `kMethodList` to respect the constant naming convention.

ddacw: It's 
```for (const libwebp::Thumbnailer::Method& method :
       {libwebp::Thumbnailer::kEqualQuality, libwebp::Thumbnailer::kEqualPSNR,
        libwebp::Thumbnailer::kSlopeOptim}) {
```
which I think is kind of messy. I'll rename it as a constant.

PR #26, opened 2020-08-26
Reduce WebPEncode() calls.

uyennguyen0903:
- Add an array of pair (size, psnr) for each frame to store encoded sizes and psnr when call WebPEncode() for different quality values.
- Avoid call WebPEncode() for the same quality several times.

### DIFF
--- a/src/thumbnailer.cc
+++ b/src/thumbnailer.cc
@@ -56,19 +56,25 @@ Thumbnailer::Status Thumbnailer::AddFrame(const WebPPicture& pic,
   return kOk;
 }
 
-Thumbnailer::Status Thumbnailer::GetPictureStats(const WebPPicture& pic,
-                                                 const WebPConfig& config,
-                                                 int* const pic_size,
+Thumbnailer::Status Thumbnailer::GetPictureStats(int ind, int* const pic_size,
                                                  float* const pic_PSNR) {
+  int quality = int(frames_[ind].config.quality);
+  if (!frames_[ind].config.near_lossless &&
+      frames_[ind].lossy_data[quality].first != -1) {
+    *pic_size = frames_[ind].lossy_data->first;
+    *pic_PSNR = frames_[ind].lossy_data->second;
+    return kOk;
+  }
+
   WebPPicture new_pic;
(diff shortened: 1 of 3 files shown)


src/thumbnailer.cc
```diff
@@ -381,6 +402,7 @@ Thumbnailer::Status Thumbnailer::TryNearLossless(WebPData* const webp_data) {
       return kMemoryError;
     }
     WebPPictureFree(&new_pic);
+    curr_ind++;
```
miwucs: small detail: the style guide requires using ++curr_ind by default (that is, unless you need the semantic of cur_ind++ like you are doing in some other places)
https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

src/thumbnailer.h
```diff
@@ -99,6 +99,11 @@ class Thumbnailer {
     int final_quality = -1;
     float final_psnr;
     bool near_lossless = false;
+    std::pair<int, float>
+        lossy_data[101];  // Array containing pairs of (size, psnr) for
+                          // qualities in range [0..100] when using lossy
+                          // encoding. If WebPEncode() has not been called for
+                          // quality 'x', lossy_data['x'] = (-1,-1.0).
```
miwucs: minor: put the comment above the variable, I think it will be shorter/easier to read this way.

src/thumbnailer.h
```diff
@@ -107,11 +112,9 @@ class Thumbnailer {
   int byte_budget_;
   int minimum_lossy_quality_;
 
-  // Compute the size (in bytes) and PSNR of a re-encoded WebPPicture at given
-  // config. The resulting size and PSNR will be stored in '*pic_size' and
-  // '*pic_PSNR' respectively.
-  Status GetPictureStats(const WebPPicture& pic, const WebPConfig& config,
-                         int* const pic_size, float* const pic_PSNR);
+  // Compute the size (in bytes) and PSNR of the 'ind'-th frame. The resulting
```
miwucs: "Computes" (method description should be in third person)

miwucs (approved):
Nice! Just minor style comments.

PR #28, opened 2020-08-31
General fixes.

ddacw:
- Remove redundant debug print.
- Default print to `stderr`.
- Fix timestamps matching.
- Catch status errors.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -19,7 +19,7 @@
 #include "thumbnailer.h"
 #include "thumbnailer.pb.h"
 
-// Returns true on success or false on failure.
+// Returns true on success and false on failure.
 static bool ReadImage(const char filename[], WebPPicture* const pic) {
   const uint8_t* data = NULL;
   size_t data_size = 0;
@@ -51,8 +51,13 @@ int main(int argc, char* argv[]) {
     WebPPicture* current_frame = pics.back().get();
     WebPPictureInit(current_frame);
 
-    ReadImage(filename_str.c_str(), current_frame);
-    thumbnailer.AddFrame(*current_frame, timestamp_ms);
+    if (!ReadImage(filename_str.c_str(), current_frame)) {
+      std::cerr << "Failed to read image " << filename_str << std::endl;
+    }
+    if (thumbnailer.AddFrame(*current_frame, timestamp_ms) !=
+        libwebp::Thumbnailer::Status::kOk) {
+      std::cerr << "Failed to add frame " << filename_str << std::endl;
+    }
   }
 
   // Write animation to file.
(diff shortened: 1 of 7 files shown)


src/main.cc
```diff
@@ -63,44 +76,51 @@ int main(int argc, char* argv[]) {
...
+    } else if (!strcmp(argv[c], "-slope_opt")) {
+      // Generate animation with slope optimization, ignore 'try_equal_psnr'
+      // and 'try_near_lossless'.
       slope_optim = true;
+    } else if (!strcmp(argv[c], "-exp")) {
+      experiment = true;
     }
   }
 
+  libwebp::Thumbnailer::Status ok;
```
miwucs: "status" is a better name I think.

src/utils/thumbnailer_utils.cc
```diff
@@ -166,6 +166,9 @@ void PrintThumbnailStatsPSNR(const ThumbnailStatsPSNR& stats) {
             << "Median PSNR: " << stats.median_psnr << std::endl;
   std::cerr << std::resetiosflags(std::ios::fixed)
             << std::resetiosflags(std::ios::showpoint) << std::endl;
+
+  std::cout << stats.min_psnr << ' ' << stats.max_psnr << ' ' << stats.mean_psnr
```
miwucs, line 170: Why does this get printed to cout while the rest is on cerr? (same below)

src/thumbnailer_near_lossless.cc
```diff
@@ -174,7 +181,7 @@ Thumbnailer::Status Thumbnailer::NearLosslessEqual(WebPData* const webp_data) {
     }
   }
 
-  std::cout << "Final near-lossless pre-processing value: " << final_near_ll
+  std::cerr << "Final near-lossless pre-processing value: " << final_near_ll
```
miwucs, line 184: This is not an error so why do you think it's better to default to cerr instead of cout? In any case this kind of debug logging from within a function should be turned off completely by default, an a verbose flag could be added to turn it on. But that can be in another PR.

miwucs (approved):

src/thumbnailer_near_lossless.cc
ddacw: I tried a `-verbose` option before removing all `EqualPSNR` debugs. I think the `stderr` is minimal enough for now.

src/thumbnailer.h
```diff
@@ -49,17 +49,17 @@ class Thumbnailer {
...
-    kMemoryError,       // In case of memory error.
-    kImageFormatError,  // If frame dimensions are mismatched.
-    kByteBudgetError,   // If there is no quality that makes the animation fit
-                        // the byte budget.
-    kStatsError,        // In case of error while getting frame's size and PSNR.
-    kWebPMuxError,      // In case of error related to WebPMux object.
-    kSlopeOptimError,   // In case of error while using slope optimization to
-                        // generate animation.
-    kGenericError       // For other errors.
+  enum [[nodiscard]] Status{
```
miwucs, line 52: yay for [[nodiscard]]!

miwucs (approved):
Sorry for the delay!

PR #29, opened 2020-09-02
Improve thumbnailer_slope_optim.cc

uyennguyen0903:


src/thumbnailer_slope_optim.cc
```diff
@@ -27,13 +27,26 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationSlopeOptim(
   CHECK_THUMBNAILER_STATUS(LossyEncodeSlopeOptim(webp_data));
   CHECK_THUMBNAILER_STATUS(NearLosslessEqual(webp_data));
 
+  // If there is no frame encoded in near-lossless, use
+  // GenerateAnimationEqualQuality() to generate animation.
+  int num_near_ll_frames = 0;
+  for (auto& frame : frames_)
```
y-guyon: Please use curling braces for the `for` loop here. They may be omitted for the `if` block, however at WebP we only skip them if both the condition and its statement fit on the same line, such as:
```c++
for (auto& frame : frames_) {
  if (frame.near_lossless) ++num_near_ll_frames;
}
```
https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements

Also if the type is not very long, try to avoid `auto`. Here the `frame` can be `const` also.

src/thumbnailer_slope_optim.cc
```diff
@@ -27,13 +27,26 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationSlopeOptim(
...
+    if (frame.near_lossless) {
+      ++num_near_ll_frames;
+    }
+  if (!num_near_ll_frames) {
```
y-guyon: `if (num_near_ll_frames == 0)` is clearer to me

src/thumbnailer_slope_optim.cc
```diff
@@ -27,13 +27,26 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationSlopeOptim(
   CHECK_THUMBNAILER_STATUS(LossyEncodeSlopeOptim(webp_data));
   CHECK_THUMBNAILER_STATUS(NearLosslessEqual(webp_data));
 
+  // If there is no frame encoded in near-lossless, use
+  // GenerateAnimationEqualQuality() to generate animation.
```
y-guyon: I do not understand why the SlopeOptim is discarded if there is no near-lossless frame. It is a lossy method, right?

src/thumbnailer_slope_optim.cc
```diff
@@ -177,7 +194,7 @@ Thumbnailer::Status Thumbnailer::LossyEncodeNoSlopeOptim(
     WebPData* const webp_data) {
   int anim_size = GetAnimationSize(webp_data);
 
-  // if the `anim_size` exceed the `byte_budget`, keep the webp_data generated
+  // If the 'anim_size' exceed the 'byte_budget', keep the webp_data generated
```
y-guyon: exceeds

src/thumbnailer_slope_optim.cc
```diff
@@ -250,4 +267,80 @@ Thumbnailer::Status Thumbnailer::LossyEncodeNoSlopeOptim(
...
+          CHECK_THUMBNAILER_STATUS(
+              GetPictureStats(curr_ind, &frames_[curr_ind].encoded_size,
+                              &frames_[curr_ind].final_psnr));
+        }
+      }
+    } else {
+      break;
+    }
+
+    encoding_order.erase(encoding_order.begin());
```
y-guyon, line 320: For clarity, add a comment such as `// Discard flattest slopes to iterate on steepest ones.`

src/thumbnailer_slope_optim.cc
```diff
@@ -250,4 +267,80 @@ Thumbnailer::Status Thumbnailer::LossyEncodeNoSlopeOptim(
...
+    --num_remaining_frames;
+  }
+
+  std::cerr << "(Final quality, Near-lossless) :" << std::endl;
```
y-guyon, line 323: Still no logging feature?

src/thumbnailer_slope_optim.cc
```diff
@@ -250,4 +267,80 @@ Thumbnailer::Status Thumbnailer::LossyEncodeNoSlopeOptim(
...
+      encoding_order.push_back(std::make_pair(slope, i));
+    }
+  std::sort(encoding_order.begin(), encoding_order.end(),
+            [](const std::pair<float, int>& x, const std::pair<float, int>& y) {
+              return x.first < y.first;
+            });
+
+  WebPData new_webp_data;
+  WebPDataInit(&new_webp_data);
+  int num_remaining_frames = encoding_order.size();
```
y-guyon: `num_remaining_frames` is equal to `encoding_order.size()` all the time so it could be removed.

src/thumbnailer_slope_optim.cc
uyennguyen0903: If there's no near-lossless frame, calling `GenerateAnimationEqualQuality()` is to use lossy encoding with the same quality and also to save time rather than calling `LossyEncodeNoSlopeOptim` and then `ExtraLossyEncode`.

src/thumbnailer_slope_optim.cc
uyennguyen0903: I am not sure what "logging feature" actually is here. Could you explain more precisely please ?

y-guyon: We talked about it during the daily meetings: currently functions always output debugging information (and to `std::cerr` which should be kept for error messages only). It would be cleaner to have them disabled by default. It can be enabled either with a compile flag (something like `#if THUMBNAILER_DEBUG`) or through a binary flag (`-verbose`), up to you. In another PR though.

src/thumbnailer_slope_optim.cc
y-guyon: So we have:
```c++
LossyEncodeSlopeOptim();
NearLosslessEqual();
if (at_least_one_frame_is_near_lossless) {
  LossyEncodeNoSlopeOptim();
  ExtraLossyEncode();
} else {
  GenerateAnimationEqualQuality();
}
```
Are you saying `GenerateAnimationEqualQuality()` has the same output as `LossyEncodeNoSlopeOptim()+ExtraLossyEncode()` only if there is no near-lossless?

uyennguyen0903: Yes, that's right.

y-guyon: But the `encoding_order` stuff never happens in `GenerateAnimationEqualQuality()`. It happens in `ExtraLossyEncode()` only if `NearLosslessEqual()` frees some budget?

Why not stop at `LossyEncodeSlopeOptim()` when there is no near-lossless then? Does it outperform `GenerateAnimationEqualQuality()` only when there is at least one near-lossless frame? I am bit confused.

uyennguyen0903: In `LossyEncodeSlopeOptim()`, the qualities of all frames are reduced a bit to have more budget for `NearLosslessEqual()`, so if there's no near-lossless frame after using `NearLosslessEqual()`, generating animation with  `GenerateAnimationEqualQuality()` is the best choice in my opinion. In fact, `ExtraLossyEncoding` is used to get the highest qualities that are as close as possible to the resulting quality of `GenerateAnimationEqualQuality()`.

src/thumbnailer_slope_optim.cc
uyennguyen0903: This will be resolved in another PR.

src/thumbnailer_slope_optim.cc
y-guyon: Where is that "reduced a bit" part in `LossyEncodeSlopeOptim()`? I do not see it. Also is not `LossyEncodeNoSlopeOptim()` supposed to fill any budget gap left?

If I understood correctly, `ExtraLossyEncode()` is doing the same as `GenerateAnimationEqualQuality()` but several times on a smaller and smaller subset of frames. Thus I have trouble seeing why/when the latter is better than the former, with or without near-lossless frames (which are excluded from the binary search in both functions right?).

Anyway if you saw better results keep the code like that, but I am afraid that defaulting to `GenerateAnimationEqualQuality()` whenever there is no near-lossless (which should be the case quite a lot because the majority of animations is camera footage not fit for near-lossless) will make the slope optim barely usable. Also if there is a lot of near-lossless frames, lossy optims are also less valuable. The convenient spot for slope optim would be the right amount of near-lossless and lossy frames, which seems often unlikely.

src/thumbnailer_slope_optim.cc
```diff
@@ -30,23 +30,21 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationSlopeOptim(
...
+    if (frame.near_lossless) ++num_near_ll_frames;
   }
-
-  int curr_anim_size = webp_data->size;
-  const int KMaxIter = 5;
-  for (int i = 0; i < KMaxIter; ++i) {
-    CHECK_THUMBNAILER_STATUS(LossyEncodeNoSlopeOptim(webp_data));
-    if (curr_anim_size == webp_data->size) break;
-    curr_anim_size = webp_data->size;
+  if (num_near_ll_frames == 0) {
```
y-guyon: If you keep the code like this, please add a comment here to clearly explain this choice. Apart from that the PR looks good to me.

y-guyon (approved):

src/thumbnailer_slope_optim.cc
uyennguyen0903: `LossyEncodeSlopeOptim()` move the quality to the left in cases the RD-curves are flat, so that it reduces the animation size. 

`LossyEncodeNoSlopeOptim()` can fill the unused budget but it is not really efficiently (as I mentioned before) so that's why I implemented `ExtraLossyEncode()`.

It is true that `ExtraLossyEncode()` does the same job as `GenerateAnimationEqualQuality()` but several times. So I set default to use `GenerateAnimationEqualQuality()` whenever there is no near-lossless frame just to avoid using `LossyEncodeNoSlopeOptim()` and then `ExtraLossyEncode()`. I tested and it seems like both methods have the same the results whenever there is no near-lossless frame, so for now I think I gonna remove `GenerateAnimationEqualQuality()` here in order to avoid misunderstanding, is it ok ?

y-guyon: Yes, it is clearer, thanks!

y-guyon (approved):

PR #33, opened 2020-09-10
Fix PSNR issue for near-lossless & Speed-up thumbnail generation.

uyennguyen0903:
- Compute the correct distortion for near-lossless by decoding the encoded bitstream.
- Speed-up the `NearLosslessEqual()` and `NearLosslessDiff()` by reducing the binary search range.

### DIFF
--- a/src/thumbnailer.cc
+++ b/src/thumbnailer.cc
@@ -79,25 +79,65 @@ Thumbnailer::Status Thumbnailer::GetPictureStats(int ind, int* const pic_size,
     return kOk;
   }
 
-  WebPPicture new_pic;
-  if (!WebPPictureCopy(&frames_[ind].pic, &new_pic)) {
-    WebPPictureFree(&new_pic);
+  WebPPicture original_pic;
+  WebPPicture encoded_pic;
+  WebPMemoryWriter memory_writer;
+  WebPMemoryWriterInit(&memory_writer);
+
+  if (!WebPPictureCopy(&frames_[ind].pic, &encoded_pic)) {
+    WebPPictureFree(&encoded_pic);
     return kStatsError;
+  }
+
+  if (frames_[ind].config.lossless) {
+    encoded_pic.writer = WebPMemoryWrite;
+    encoded_pic.custom_ptr = (void*)&memory_writer;
+  } else {
+    if (!WebPPictureCopy(&frames_[ind].pic, &original_pic)) {
+      WebPPictureFree(&original_pic);
+      return kStatsError;
+    }
   }
 
   WebPAuxStats stats;
-  new_pic.stats = &stats;
-  if (!WebPEncode(&frames_[ind].config, &new_pic)) {
-    WebPPictureFree(&new_pic);
+  encoded_pic.stats = &stats;
+  if (!WebPEncode(&frames_[ind].config, &encoded_pic)) {
+    WebPPictureFree(&encoded_pic);
     return kStatsError;
   }
 
-  *pic_size = new_pic.stats->coded_size;
+  if (frames_[ind].config.lossless) {
+    if (frames_[ind].config.near_lossless == 100) {
+      // Pure lossless: image was not modified, make 'original_picture' a view
+      // of 'picture' by copying all members except the freeable pointers.
+      original_pic = encoded_pic;
+      original_pic.memory_ = original_pic.memory_argb_ = NULL;
+    } else {
+      // Decode the bitstream stored in 'memory_writer' to get the altered image
+      // to 'picture'; save the 'original_picture' beforehand.
+      original_pic = encoded_pic;
+      if (!WebPPictureInit(&encoded_pic)) {  // Do not free 'picture'.
+        return kStatsError;
+      }
+
+      encoded_pic.use_argb = 1;
+      if (!ReadWebP(memory_writer.mem, memory_writer.size, &encoded_pic,
+                    /*keep_alpha=*/WebPPictureHasTransparency(&encoded_pic),
+                    /*metadata=*/NULL)) {
+        return kStatsError;
+      }
+      encoded_pic.stats = original_pic.stats;
+    }
+    original_pic.stats = NULL;
+  }
+
+  *pic_size = encoded_pic.stats->coded_size;
 
   float distortion_result[5];
-  if (!WebPPictureDistortion(&frames_[ind].pic, &new_pic, 0,
+  if (!WebPPictureDistortion(&original_pic, &encoded_pic, 0,
                              distortion_result)) {
-    WebPPictureFree(&new_pic);
+    WebPPictureFree(&encoded_pic);
+    WebPPictureFree(&original_pic);
     return kStatsError;
   } else {
     *pic_PSNR = distortion_result[4];  // PSNR-all.
@@ -106,7 +146,10 @@ Thumbnailer::Status Thumbnailer::GetPictureStats(int ind, int* const pic_size,
   if (!frames_[ind].config.lossless) {
     frames_[ind].lossy_data[quality] = std::make_pair(*pic_size, *pic_PSNR);
   }
-  WebPPictureFree(&new_pic);
+
+  WebPPictureFree(&original_pic);
+  WebPPictureFree(&encoded_pic);
+  WebPMemoryWriterClear(&memory_writer);
 
   return kOk;
 }
--- a/src/thumbnailer.cc
+++ b/src/thumbnailer.cc
@@ -79,7 +79,6 @@ Thumbnailer::Status Thumbnailer::GetPictureStats(int ind, int* const pic_size,
     return kOk;
   }
 
-  WebPPicture original_pic;
   WebPPicture encoded_pic;
   WebPMemoryWriter memory_writer;
   WebPMemoryWriterInit(&memory_writer);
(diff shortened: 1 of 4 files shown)


src/thumbnailer.cc
```diff
@@ -79,25 +79,68 @@ Thumbnailer::Status Thumbnailer::GetPictureStats(int ind, int* const pic_size,
...
+        return kStatsError;
+      }
+
+      encoded_pic.use_argb = 1;
+      if (!ReadWebP(memory_writer.mem, memory_writer.size, &encoded_pic,
+                    /*keep_alpha=*/WebPPictureHasTransparency(&encoded_pic),
+                    /*metadata=*/NULL)) {
+        return kStatsError;
+      }
+      encoded_pic.stats = original_pic.stats;
```
miwucs, line 130: I don't understand this line. can you explain why we copy the stats from the original pic?

src/thumbnailer.cc
```diff
@@ -79,25 +79,68 @@ Thumbnailer::Status Thumbnailer::GetPictureStats(int ind, int* const pic_size,
...
   }
 
+  // Lossy will modify the 'encoded_pic' but not lossless and near-lossless.
+  // Therefore, keep the encoded bitstream in the memory and decode it to
+  // compute PSNR correctly for near-lossless.
+  if (frames_[ind].config.lossless) {
+    encoded_pic.writer = WebPMemoryWrite;
+    encoded_pic.custom_ptr = (void*)&memory_writer;
+  } else {
+    if (!WebPPictureCopy(&frames_[ind].pic, &original_pic)) {
```
miwucs: I'm not sure why you need to copy the frame to original_pic. Can't original_pic just be a view on frames_[ind].pic since I don't think that one ever gets modified?

src/thumbnailer.cc
```diff
@@ -79,25 +79,68 @@ Thumbnailer::Status Thumbnailer::GetPictureStats(int ind, int* const pic_size,
...
+  encoded_pic.stats = &stats;
+  if (!WebPEncode(&frames_[ind].config, &encoded_pic)) {
+    WebPPictureFree(&encoded_pic);
     return kStatsError;
   }
 
-  *pic_size = new_pic.stats->coded_size;
+  if (frames_[ind].config.lossless) {
+    if (frames_[ind].config.near_lossless == 100) {
+      // Pure lossless: image was not modified, make 'original_pic' a view
```
miwucs: we could even skip computing the distortion in this case as it should be 99, but maybe not super important

src/thumbnailer.cc
```diff
@@ -79,25 +79,68 @@ Thumbnailer::Status Thumbnailer::GetPictureStats(int ind, int* const pic_size,
...
+
+  *pic_size = encoded_pic.stats->coded_size;
 
   float distortion_result[5];
-  if (!WebPPictureDistortion(&frames_[ind].pic, &new_pic, 0,
+  if (!WebPPictureDistortion(&original_pic, &encoded_pic, 0,
                              distortion_result)) {
-    WebPPictureFree(&new_pic);
+    WebPPictureFree(&encoded_pic);
+    WebPPictureFree(&original_pic);
```
miwucs: do you need WebPMemoryWriterClear(&memory_writer); here as well?

src/thumbnailer_near_lossless.cc
```diff
@@ -24,8 +24,14 @@ Thumbnailer::Status Thumbnailer::NearLosslessDiff(WebPData* const webp_data) {
...
     float curr_psnr = frame.final_psnr;
 
-    int min_near_ll = 0;
-    int max_near_ll = 100;
+    // Binary search for near-lossless with the pre-processing values contained
+    // in the 'preprocessing_list' to speed-up the algorithm. It is not
+    // necessary to binary search for all the values in range [0,100]
+    // since actually for near-lossless, the frame size and PSNR are not changed
+    // when the preprocessing increases a small quantity.
+    const int preprocessing_list[6] = {0, 20, 40, 60, 80, 100};
```
miwucs: Can you make this a file constant? Then you can reuse it in NearLosslessEqual, using min_ind=1

src/thumbnailer_near_lossless.cc
```diff
@@ -186,32 +199,36 @@ Thumbnailer::Status Thumbnailer::NearLosslessEqual(WebPData* const webp_data) {
...
+  }
+
+  if (final_near_ll != 0) {
+    CHECK_THUMBNAILER_STATUS(GenerateAnimationNoBudget(&new_webp_data));
+    if (new_webp_data.size <= byte_budget_) {
+      WebPDataClear(webp_data);
+      *webp_data = new_webp_data;
     } else {
-      max_near_ll = mid_near_lossless - 1;
+      WebPDataClear(&new_webp_data);
```
miwucs, line 221: if we discard the last encode attempt, does that mean the final result is not in sync with the settings in frames_? Is that ok?

src/thumbnailer.cc
uyennguyen0903: I copied this line from `cwebp`. I think at this [line](https://github.com/googleinterns/step255-2020/blob/6ece994afa585b12d255dd7ab8bb2e592d9887fa/src/thumbnailer.cc#L122) the `encoded_pic.stats` is probably modified, I take a look into it, if it is not necessary I will remove it.

src/thumbnailer.cc
uyennguyen0903: Yeah, it sounds reasonable I will remove `original_pic`.

src/thumbnailer_near_lossless.cc
uyennguyen0903: You are right, the final result here must be synchronized with preprocessing 0 if `new_webp_data.size` exceeds the budget, I will fix it!

src/thumbnailer.cc
uyennguyen0903: I checked this line. The `encoded_pic.stats` is NULL here, so we need to copy the stats from `original_pic`.

src/thumbnailer.cc
```diff
@@ -95,30 +94,29 @@ Thumbnailer::Status Thumbnailer::GetPictureStats(int ind, int* const pic_size,
...
+      *pic_size = encoded_pic.stats->coded_size;
+      WebPPictureFree(&encoded_pic);
+      WebPMemoryWriterClear(&memory_writer);
+      return kOk;
     } else {
       // Decode the bitstream stored in 'memory_writer' to get the altered
-      // image, save the 'original_pic' beforehand.
-      original_pic = encoded_pic;
+      // image.
+      WebPPicture original_pic = encoded_pic;
```
uyennguyen0903, line 119: I made a copy of `encoded_pic` to a new `WebPPicture` here because if not there will be memory leaks. Also, if I write `frames_[ind].pic = encoded_pic;` there will be segmentation fault.

src/thumbnailer_near_lossless.cc
```diff
@@ -14,6 +14,12 @@
 
 #include "thumbnailer.h"
 
+// List of pre-processing values used in binary search for near-lossless to
+// speed-up the algorithm. It is not necessary to binary search for all the
+// values in range [0,100] since actually for near-lossless, the frame size and
+// PSNR are not changed when the preprocessing increases a small quantity.
+const int preprocessing_list[6] = {0, 20, 40, 60, 80, 100};
```
miwucs: rename it kPreprocessingList to follow the styleguide
also put it in an anonymous namespace inside the libwebp namespace:
namespace libwebp {
namespace {
  const int preprocessing_list[6] = {0, 20, 40, 60, 80, 100};
}  // namespace

This is to avoid possible conflicts with constants in other files, even if it's fairly unlikely.

src/thumbnailer_near_lossless.cc
```diff
@@ -222,6 +216,12 @@ Thumbnailer::Status Thumbnailer::NearLosslessEqual(WebPData* const webp_data) {
       *webp_data = new_webp_data;
     } else {
       WebPDataClear(&new_webp_data);
+      for (int curr_ind : near_ll_frames) {
+        frames_[curr_ind].config.near_lossless = 0;
+        CHECK_THUMBNAILER_STATUS(
+            GetPictureStats(curr_ind, &frames_[curr_ind].encoded_size,
```
miwucs: This re encodes all the frames, which is extra work, right? Maybe we can save the previous stats instead and restore them? (you can add a TODO comment to fix this in another commit if you want)

miwucs (approved):

src/thumbnailer_near_lossless.cc
uyennguyen0903: I resolved this issue in this PR too because it it not complicated. 
PTAL.

src/thumbnailer_near_lossless.cc
```diff
@@ -14,13 +14,12 @@
...
 
+namespace libwebp {
 // List of pre-processing values used in binary search for near-lossless to
 // speed-up the algorithm. It is not necessary to binary search for all the
 // values in range [0,100] since actually for near-lossless, the frame size and
 // PSNR are not changed when the preprocessing increases a small quantity.
-const int preprocessing_list[6] = {0, 20, 40, 60, 80, 100};
-
-namespace libwebp {
+const int kPreprocessingList[6] = {0, 20, 40, 60, 80, 100};
```
miwucs: it's better practice to also put this in an anonymous namespace, or you can add the "static" keyword before "const",  which has a similar effect

miwucs (approved):

PR #34, opened 2020-09-10
Add bazel build.

ddacw:


test/BUILD
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_cc//cc:defs.bzl", "cc_binary")
+
+cc_binary(
+    name = "thumbnailer-test",
```
miwucs: thumbnailer_test since you use underscores everywhere else

examples/BUILD
```diff
@@ -0,0 +1,32 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
+    name = "example_util",
+    srcs = ["example_util.c"],
+    hdrs = ["example_util.h"],
+    visibility = [
```
miwucs: I think that for such a small project it would be ok to make everything public but up to you

miwucs (approved):
nice work!

PR #35, opened 2020-09-11
Simplify API.

uyennguyen0903:


test/thumbnailer_test.cc
```diff
@@ -103,15 +103,14 @@ class WebPTestGenerator {
   }
 };
 
-class GenerateAnimationTest : public ::testing::TestWithParam<
-                                  std::tuple<int, uint8_t, bool, bool, int>> {};
+class GenerateAnimationTest
+    : public ::testing::TestWithParam<std::tuple<int, uint8_t, bool, int>> {};
```
y-guyon: If you used a `Method` directly instead of an `int`, you might be able to use `::testing::ValuesIn(kMethodList)` instead of `::testing::Values(0, 1, 2, 3, 4)`

y-guyon (approved):

PR #36, opened 2020-09-13
Add Abseil flags. Refactor.

ddacw:
### In this PR:
- Add flags.
- Use `unsigned`, `size_t`, `std::size_t`, `size_type` where needed (to fix bazel warnings).
- Move utilities (e.g. `ReadImage`) to `src/utils`.
- Fix memory leak in `thumbnailer` binary with `libwebp::WebPPictureDelete`.


src/main.cc
```diff
@@ -12,94 +12,116 @@
...
-
-  return ok;
+#include "utils/thumbnailer_utils.h"
+
+ABSL_FLAG(std::string, o, "out.webp", "Output file name.");
+
+// Thumbnailer options.
+ABSL_FLAG(uint32_t, soft_max_size, 153600,
+          "Desired (soft) maximum size limit in bytes.");
+ABSL_FLAG(uint32_t, loop_count, 0, "Number of times animation will loop.");
```
y-guyon: " (zero for infinity)"

src/main.cc
```diff
@@ -12,94 +12,116 @@
...
+ABSL_FLAG(uint32_t, min_lossy_quality, 0,
+          "Minimum lossy quality to be used for encoding each frame.");
+ABSL_FLAG(uint32_t, hard_max_size, 153600,
```
y-guyon, line 32: It would make more sense to put `hard_max_size` next to `soft_max_size` no? Try to group flags by meaning. For example: webp encoding options (loop_count etc.), then thumbnailer algorithm options (max_size, slope stuff etc.), then binary options (verbose etc.)

src/main.cc
```diff
@@ -12,94 +12,116 @@
...
 int main(int argc, char* argv[]) {
   GOOGLE_PROTOBUF_VERIFY_VERSION;
-  thumbnailer::ThumbnailerOption thumbnailer_option;
 
-  // Method generating thumbnail, the default method is using lossy encode and
-  // imposing the same quality to all frames.
-  libwebp::Thumbnailer::Method method =
-      libwebp::Thumbnailer::Method::kEqualQuality;
+  absl::SetProgramUsageMessage(absl::StrFormat(
+      "Usage: %s [options] %s -o=output.webp", argv[0], "frame_list.txt"));
```
y-guyon: Why not write directly `frame_list.txt` instead of `%s`

src/main.cc
```diff
@@ -12,94 +12,116 @@
...
   }
 
+  // Initialize thumbnailer.
   libwebp::Thumbnailer thumbnailer = libwebp::Thumbnailer(thumbnailer_option);
 
   // Process list of images and timestamps.
+  int input_list_c = 0;
+  for (int c = 1; c < argc; ++c) {
+    if (argv[c][0] != '-') {
+      input_list_c = c;
```
y-guyon: `break;` here otherwise it might keep the wrong arg such as "123" with "thumbnailer frame_list.txt -loop_count 123" and this only valid if `frame_list.txt` is the first argument (but then why parse all `argc`?) or if `=` are used in flags (can it be enforced?)

src/main.cc
```diff
@@ -109,23 +131,36 @@ int main(int argc, char* argv[]) {
     return 1;
   }
 
-  // Generate animation.
-  const char* output = argv[2];
+  // Generate the animation.
+  const std::string& output = absl::GetFlag(FLAGS_o);
```
y-guyon: Did you check that `absl::GetFlag()` returned a reference `string&` and not a value `string`?

src/main.cc
```diff
@@ -109,23 +131,36 @@ int main(int argc, char* argv[]) {
...
+
+  if (absl::GetFlag(FLAGS_equal_psnr)) {
+    method = libwebp::Thumbnailer::Method::kEqualPSNR;
+  } else if (absl::GetFlag(FLAGS_near_ll_diff)) {
+    method = libwebp::Thumbnailer::Method::kNearllDiff;
+  } else if (absl::GetFlag(FLAGS_near_ll_equal)) {
+    method = libwebp::Thumbnailer::Method::kNearllEqual;
+  } else if (absl::GetFlag(FLAGS_slope_optim)) {
+    method = libwebp::Thumbnailer::Method::kSlopeOptim;
+  }
```
y-guyon, line 149: I wonder if an error or a warning should be printed in case several methods are passed in flags, such as "-equal_psnr -slope_optim" which will only use `EqualPSNR` although `SlopeOptim` was specified after and thus should override

src/thumbnailer.h
```diff
@@ -90,9 +90,9 @@ class Thumbnailer {
     WebPPicture pic;
     int timestamp_ms;  // Ending timestamp in milliseconds.
     WebPConfig config;
-    int encoded_size;
+    size_t encoded_size = 0;
```
y-guyon, line 88: Unclear why some fields are initialized (`encoded_size` etc.) and not others (`timestamp_ms`)

src/main.cc
```diff
@@ -12,94 +12,116 @@
...
+          "pre-processing value for all near-lossless frames.");
+ABSL_FLAG(bool, slope_optim, false,
+          "Generate animation with slope optimization.");
+
+// Returns false on invalid configurations.
+bool ThumbnailerValidateOption(
+    const thumbnailer::ThumbnailerOption& thumbnailer_option) {
+  if (thumbnailer_option.min_lossy_quality() > 100) return false;
+  if (thumbnailer_option.webp_method() > 6) return false;
+  if (thumbnailer_option.slope_dpsnr() < 0) return false;
```
y-guyon, line 67: `> 99` check?

src/utils/thumbnailer_utils.cc
```diff
@@ -75,7 +99,7 @@ UtilsStatus AnimData2PSNR(const std::vector<Frame>& original_frames,
     return kGenericError;
   }
 
-  int new_frame_index = 0;
+  std::vector<Frame>::size_type new_frame_index = 0;
```
y-guyon: `size_t` or `uint32_t` should be fine

src/utils/thumbnailer_utils.cc
```diff
@@ -159,7 +182,7 @@ void PrintThumbnailStatsPSNR(const ThumbnailStatsPSNR& stats,
   std::cout << std::fixed << std::showpoint;
   std::cout << std::setprecision(3);
 
-  for (int i = 0; i < stats.psnr.size(); ++i) {
+  for (std::vector<float>::size_type i = 0; i < stats.psnr.size(); ++i) {
```
y-guyon: `size_t` should be enough

src/main.cc
ddacw: For abseil, I think `=` is used by default. The filename should be at the end of the argument list (as @miwucs said) so I take the last valid args.

src/main.cc
ddacw: I think it would be better if there's a proto message for each. For now I just rearranged the options and added comments in `main.cc`.

src/main.cc
ddacw: I'm not sure if absl::Flags can handle this, since it would be awkward checking manually.

src/main.cc
```diff
@@ -95,19 +101,13 @@ int main(int argc, char* argv[]) {
...
 
   // Process list of images and timestamps.
-  int input_list_c = 0;
-  for (int c = 1; c < argc; ++c) {
-    if (argv[c][0] != '-') {
-      input_list_c = c;
-    }
-  }
-  if (input_list_c == 0) {
+  if (positional_args.empty()) {
```
y-guyon: `if (positional_args.size() != 1) {` unless several positional arguments are accepted

src/main.cc
y-guyon: The four `ABSL_FLAG(bool)` could be merged into a single `ABSL_FLAG(std::string)` so that the last method arg is used. I do not know if this is better though.

y-guyon (approved):

PR #38, opened 2020-09-15
Refactor slope_optim.

uyennguyen0903:
- Refactor thumbnailer_slope_optim.cc (remove the `new_optim_list` vector from `LossyEncodeSlopeOptim()` to avoid misunderstanding).
- Add flag `-equal_quality` to `main.cc`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ bazel build "//src:thumbnailer" && bazel build "//src/utils:thumbnailer_compare"
 
 This tool reads a sequence of images and timestamps from a text file and produces an animation in WebP format.
 
-**Usage:**
+#### Usage
 
 ```
 ./bazel-bin/src/thumbnailer [options] frame_list.txt -o=output.webp
@@ -35,7 +35,7 @@ This tool reads a sequence of images and timestamps from a text file and produce
 
 By default, the lossy compression method imposing the same quality to all frames will be used to produce animation. 
 
-**Example:**
+#### Example:
 
 Command line to generate animation using various options:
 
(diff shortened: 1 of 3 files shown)


src/main.cc
```diff
@@ -47,6 +47,8 @@ ABSL_FLAG(bool, allow_mixed, false, "Use mixed lossy/lossless compression.");
 ABSL_FLAG(bool, verbose, false, "Print various encoding statistics.");
 
 // Thumbnailer methods.
+ABSL_FLAG(bool, equal_quality, false,
+          "Generate animation so that all frames have the same quality.");
```
y-guyon: What is the point of this flag if it is the default behavior?

y-guyon (approved):

src/main.cc
uyennguyen0903: I think it should be clearer to have this flag. Without this flag, if user intends to create animation with `equal_quality` method and write `-equal_quality` in command line, the thumbnailer will show error in this case.

y-guyon: Then a single `ABSL_FLAG(std::string)` might make more sense than several `ABSL_FLAG(bool)` (see [comment](https://github.com/googleinterns/thumbnailer/pull/36#discussion_r488001003)), up to you

README.md
```diff
@@ -51,29 +52,31 @@ The following example shows the format of the text file containing frame filenam
 ./anim1/frame04.png 600
 ```
 
-**WebP Encoding Options:**
+**Options:**
 
-| Option | Default Value | Description|
-|--------|:-------------:|------------|
+| <img width="150"> Option | Default Value | <img width="600"> Description|
```
y-guyon: Is there another way to center the text rather than manual padding with `` tags?

README.md
```diff
@@ -51,29 +52,31 @@ The following example shows the format of the text file containing frame filenam
...
-|`-equal_quality`|true|Generate animation setting the same quality to all frames.|
-|`-equal_psnr`|false|Generate animation so that all frames have the same PSNR.|
-|`-near_ll_diff`|false|Generate animation allowing near-lossless method, impose different pre-processing to near-losslessly-encoded frames.|
-|`-near_ll_equal`|false|Generate animation allowing near-lossless method, impose the same pre-processing to near-losslessly-encoded frames.|
-|`-slope_optim`|false|Generate animation with slope optimization.|
+|`-algorithm`|equal_quality|Algorithm to generate animation {equal_quality, equal_psnr, near_ll_diff, near_ll_equal, slope_optim}.|
 |`-slope_dpsnr`|1.0|Maximum PSNR change (in dB) used in slope optimization.|
 |`-verbose`|false|Print various encoding statistics showed by chosen algorithm.|
 
+**`-algorithm` flag description:**
```
y-guyon: You should use `## Title` here I think. See https://google.github.io/styleguide/docguide/style.html

README.md
```diff
@@ -51,29 +52,31 @@ The following example shows the format of the text file containing frame filenam
...
+
+| <img width="100"> Algorithm | <img width="600"> Description |
+|-----------------------------|-------------------------------|
+|`-equal_quality`|Generate animation setting the same quality to all frames.|
```
y-guyon: This is no longer `-equal_quality` but `-algorithm equal_quality`

src/main.cc
```diff
@@ -140,15 +129,23 @@ int main(int argc, char* argv[]) {
...
-  } else if (absl::GetFlag(FLAGS_near_ll_equal)) {
+  } else if (method_flag == "near_ll_equal") {
+    // Generate animation allowing near-lossless method. Use the same
+    // pre-processing value for all near-lossless frames.
     method = libwebp::Thumbnailer::Method::kNearllEqual;
-  } else if (absl::GetFlag(FLAGS_slope_optim)) {
+  } else if (method_flag == "slope_optim") {
+    // Generate animation with slope optimization.
     method = libwebp::Thumbnailer::Method::kSlopeOptim;
   }
```
y-guyon, line 153: `else { std::cerr << "Unknown -algorithm " << method_flag ...`

y-guyon (approved):

README.md
uyennguyen0903: I've tried several ways but it didn't work.

y-guyon: Up to you but it feels like a trick. If the purpose was to fix the broken flags split on two lines, just remove the `-` I guess

PR #40, opened 2020-09-17
Refactor slope_optim.

uyennguyen0903:
Merge `ExtraLossyEncode()` into `GenerateAnimationEqualQuality()`.

### DIFF
--- a/src/thumbnailer.cc
+++ b/src/thumbnailer.cc
@@ -250,19 +250,25 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualQuality(
               return a.timestamp_ms < b.timestamp_ms;
             });
 
-  // 'slope_optim = true' means that the slope optimization process has been
-  // already called beforehand.
-  bool slope_optim = (frames_[0].final_quality != -1);
+  // If slope optimization process has been already called beforehand, the
+  // 'slope_optim_done' is true.
+  bool slope_optim_done = (frames_[0].final_quality != -1);
 
   // Use binary search to find the quality for lossy compression that makes the
   // animation fit right below the given byte budget.
-  int min_quality = 100;
-  for (std::size_t i = 0; i < frames_.size(); ++i) {
-    if (!frames_[i].near_lossless) {
-      min_quality = std::min(min_quality, frames_[i].final_quality + 1);
+  int min_quality;
+  if (slope_optim_done) {
+    min_quality = 100;
+    for (std::size_t i = 0; i < frames_.size(); ++i) {
+      if (!frames_[i].near_lossless) {
+        min_quality = std::min(min_quality, frames_[i].final_quality + 1);
+      }
     }
+  } else {
+    min_quality = 0;
   }
   min_quality = std::max(min_quality, minimum_lossy_quality_);
+
   int max_quality = 100;
   int final_quality = -1;
   WebPData new_webp_data;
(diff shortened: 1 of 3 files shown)


src/thumbnailer.cc
```diff
@@ -250,18 +250,31 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualQuality(
               return a.timestamp_ms < b.timestamp_ms;
             });
 
-  // Use binary search to find the quality that makes the animation fit right
-  // below the given byte budget.
-  int min_quality = minimum_lossy_quality_;
+  // 'slope_optim = true' means that the slope optimization process has been
+  // already called beforehand.
+  bool slope_optim = (frames_[0].final_quality != -1);
```
y-guyon: Maybe `slope_optim_done` or similar then

src/thumbnailer.cc
```diff
@@ -250,18 +250,31 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualQuality(
...
+
+  // Use binary search to find the quality for lossy compression that makes the
+  // animation fit right below the given byte budget.
+  int min_quality = 100;
+  for (std::size_t i = 0; i < frames_.size(); ++i) {
+    if (!frames_[i].near_lossless) {
+      min_quality = std::min(min_quality, frames_[i].final_quality + 1);
+    }
+  }
```
y-guyon: Should this be done only when `slope_optim` or are all `frames_[i].final_quality` correctly set otherwise?

src/thumbnailer.cc
```diff
@@ -250,18 +250,31 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualQuality(
...
   int max_quality = 100;
   int final_quality = -1;
   WebPData new_webp_data;
   WebPDataInit(&new_webp_data);
 
   while (min_quality <= max_quality) {
     int mid_quality = (min_quality + max_quality) / 2;
-    for (auto& frame : frames_) {
-      frame.config.quality = mid_quality;
+    for (std::size_t i = 0; i < frames_.size(); ++i) {
```
y-guyon: The `for (auto& frame : frames_)` loop could have been kept

src/thumbnailer.cc
```diff
@@ -277,16 +290,19 @@ Thumbnailer::Status Thumbnailer::GenerateAnimationEqualQuality(
...
   }
 
-  int curr_ind = 0;
-  for (auto& frame : frames_) {
-    frame.final_quality = final_quality;
-    frame.config.quality = final_quality;
-    CHECK_THUMBNAILER_STATUS(
-        GetPictureStats(curr_ind++, &frame.encoded_size, &frame.final_psnr));
+  for (std::size_t i = 0; i < frames_.size(); ++i) {
+    if (!frames_[i].near_lossless && frames_[i].final_quality < final_quality) {
```
y-guyon, line 299: Should `frames_[i].final_quality > final_quality` be tested also?

y-guyon (approved):
Nice clean-up! Is there any impact on the algorithm output?

src/thumbnailer.cc
uyennguyen0903: `frames_[i].final_quality` is initialized by -1 in the header file, so `min_quality` will be 0 if `slope_optim=false`. But I could make this be clearer by adding something like:
```
if (!slope_optim) {
  min_quality = 0;
} else {
  ...
}
```
Should I keep the code or modify it in this way ?
Btw, the algorithm output was not changed but the speed for `slope_optim` was improved a lot!

y-guyon: In general, it is clearer to assert facts in order to know which data contains what and when. I would modify it as you described. The other solution would be to assert each `frames_[i].final_quality` if `slope_optim=false`.
Good!

src/thumbnailer.cc
uyennguyen0903: I think [here](https://github.com/googleinterns/thumbnailer/blob/0d899e58a3b9256da63b97dbcdecc8df4d4b49f7/src/thumbnailer.cc#L276) , there's no change if `frames_[i].final_quality > final_quality`, so it does not need to be tested.

y-guyon (approved):

